Issue #257, opened 2015-02-25
4.6 readyness


magnolia61:
With the 4.6 in beta since today i was wondering, do you foresee any issues with compatibility (or would you think this extension will have no big issues working in 4.6? Of course I will also try and test myself, but you know the architecture of the extension and changes in civicrm better than I do :-)

systopia:
@magnolia61 We currently don't have time/funding to do 4.6 compatibility. Just give it a try, and let us know your findings.

tttp:
@magnolia61 did you test on 4.6?

magnolia61:
Yep, I am testing the latest code with civicrm 4.6.2 right now. Backend functions seem to work ok.
The only thing so far I run into is a regression of the empty hidden start_date field (https://github.com/Project60/org.project60.sepa/issues/223). @nikobo fixed that for 4.5 but maybe it that fix is not compatible with 4.6. I have no clue why that issue returned.

magnolia61:
For 4.6 compatibility the work Eileen has done on payment processor integration would seem very interesting. See https://issues.civicrm.org/jira/browse/CRM-15555
Tips from Eileen: 
Your payment processor should override things like supportsBackOffice to declare whether it should show as an option on the backoffice form.
ValidatePaymentInstrument overrides form validation - ideally you should be able to implement it without using hooks now & if you are still using hooks we should ensure we have the things requiring the use of hooks on our road map

antrik:
On Thu, Apr 30, 2015 at 05:34:59AM -0700, Richard van Oosterhout wrote:
> For 4.6 compatibility the work Eileen has done on payment processor integration would seem very interesting.

This is very relevant for the SfE variant of the SEPA extension, as we are actually using the upstream back-office forms, and will hopefully need considerably less hacks now.

The systopia/project60 variant OTOH shouldn't be affected AIUI, as it implements its own backoffice forms.

magnolia61:
I dream and hope there can be a renewed effort to integrate both variants towards one common release with the best of both worlds because it seems so much energy could be synergy here... </offtopic>

systopia:
> I dream and hope there can be a renewed effort to integrate both variants towards one common release with the best of both worlds because it seems so much energy could be synergy here... #offtopic

I agree, but the merge would be pretty complex, and there is simply no funding for this atm.

magnolia61:
[...]

PR #270, opened 2015-04-30
Fix for missing contactID in 4.6

magnolia61:
The code broke in 4.6 over the folllowing line:
$params['contact_id'] = $this->getForm()->getVar('_contactID');
Changed that to:
$params['contact_id']    = $params['contactID'];


tttp:
@magnolia61 did you check if it's working also on 4.5? 
Might be safer to 
   if (!array_key_exists("contactID",$params)) 
     $params['contact_id'] = $this->getForm()->getVar('_contactID');

systopia:
@tttp is right, it needs to be working on 4.4, 4.5 and 4.6, so we can't merge this.
I'll create a fix based on @tttp's approach when we look into #257

Issue #223, opened 2015-01-06
"Please select start date"


magnolia61:
Today I did a clean install of the bleeding dev_pp code with the 4 changes as of jan 6th.
I get asked to enter the start date so something seem unstable there and the hidden field doesn;t get filled. I do not know what causes this or where to look for a solution.

I completely uninstalled civisepa, removing files, cleaning cache both from drupal and civicrm.

systopia:
@magnolia61 This is a problem with CiviCRM 4.5, which is not yet supported by that particular branch. However, we're working on this right now and it should be fixed when we merge it back into master by the end of the week.

nikobo:
This issue should be fixed for CiviCRM 4.4.x with 5b3b526e6a

magnolia61:
After todays commite in de the dev_pp branch the hidden start date functions ok.

magnolia61:
Hi @nikobo , do you remember what line actually fixed the empty hidden start_date field for 4.5. In 4.6 the field is empty again. Browsing through the code in SDD.php and SDD.tpl a cannot figure out which line is responsible for setting the hidden field to todays date. Can you help me bugtrace this one?

systopia:
@nikobo Could you verify this?

nikobo:
Sorry for the late reply. I didn't notice this ticket until now. The problem here is that quickform assigns a partially randomized name for the start_date-field. You can view a possible solution in the referenced commit/PR.

systopia:
@magnolia61 Could you please confirm with the current ``master``, that Niko's PR (https://github.com/Project60/org.project60.sepa/pull/273) works for you?

magnolia61:
i gladly confirm that the start-date is working correctly with the current master, Nikoś PR and civicrm 4.6.3 ;-)

systopia:
Great, thank you!

PR #273, opened 2015-05-20
fix for #223

nikobo:
- works with 4.5
- works with 4.6


systopia:
also seems to work with 4.4.11

PR #276, opened 2015-05-22
payment processor should not ignore cycle_day parameter

### LINKED ISSUE #274: Cycle days are not correctly considered when using SEPA-Payment processor for recurring contributions
If you enter a recurring contribution on a contribution page, the sepa-payment processor is supposed to derive the earliest possible cycle day for all payments from the creditor's batching settings (considering the notice day and cycle day settings).

However, this does not seem to work, at least for 4.6.x. I created a recurring contribution which should have gotten the 5th as a cycling day (which was also displayed on the confirmation page). However the collection day of the recurring contribution was set to the 1st.

nikobo:
This should fix #274. Compatible with 4.4.11 - 4.6.


systopia:
seems to work

Issue #281, opened 2015-06-01
Settings: creditor contact lookup doesn't work


systopia:
The CiviSEPA settings panel features a field that should do a name lookup to retrieve the creditor contact. This seems not to be working on CiviCRM 4.6.

 1. In this context, I would like to enhance the lookup, and offer an additional field next to the creditor lookup, where you can enter a contact ID. 
 2. the form should validate that a proper creditor contact ID has been set. Currently, this is not the case, and the JS error that blocks the lookup does not prevent you from saving the creditor with no contact ID.
 3. the form should feature the default organisation (can be obtained from the config singleton)

systopia:
@nikobo Could you have a look?

nikobo:
error: ```Uncaught Error: cannot call methods on autocomplete prior to initialization; attempted to call method '/CiviSEPA_4_5_8/civicrm/ajax/rest?className=CRM_Contact_Page_AJAX&fnName=getContactList&json=1&context=contact'```

I think the problem is that Jquery(UI?) has been updated and the autocomplete field has to be initialized/used differently now.

systopia:
I had to remove a line that blocks editing settings in general (see https://github.com/Project60/org.project60.sepa/commit/28b9f0aad8220742d6f4b67d3b8c9d30742bfadf). We will have to have another look at this...

systopia:
@nikobo Please test (4.4,4.5,4.6) and then close the ticket. Please pay attention to these situations:
 * edit creditor
 * create creditor
 * clear the creditor contact field and try to save => not possible
 * settings page still works w/o touching creditors

PR #287, opened 2015-06-08
fix for #281

nikobo:
tested with:
- 4.4.15
- 4.5.8
- 4.6.3

This does not yet include the additional id (integer) field


nikobo:
#281

PR #291, opened 2015-06-15
require creditor contact via js (#281)

nikobo:
With this commit you now need to set a valid creditor contact before you can save the creditor.


systopia:
@nikobo let's go through these changes together

PR #297, opened 2015-07-06
Additional file formats

scardinius:
Hi,

What do you say for such approach to supporting additional file formats? I'd like to prepare new format for Citibank and this format isn't xml but txt.

```sql
--
-- upgrade database (insert new format (txt) and modify previous (xml))
--
SET @option_group_id = (SELECT `id` FROM `civicrm_option_group` WHERE `name` = 'sepa_file_format');

INSERT INTO civicrm_option_value (option_group_id, label, value, name, grouping, filter, is_default, weight)
VALUES (@option_group_id, 'CitiBank Polecenie Zaplaty', 4, 'citibank-pl', 'txt', 0, 0, 1);

UPDATE civicrm_option_value
SET grouping = 'xml'
WHERE option_group_id = @option_group_id AND value IN (1, 2, 3);
```


tttp:
hi,

use civicrm_api3 to create new optionvalues instead of aiming directly at the db

assuming a separate extension that extends sepa, you should look at managed entities, that let the installer automatically add/remove the optionValue (or any entity) when you enable/disable the extension

michau:
@tttp How about the pr coming together with this issue? The SQL is just an additional comment. :-)

michau:
Also speaking about separate extension - our aim is to add new format for bank data exchange: an extension for extension seems a bit like an overkill, wouldn't it be easier to have "abstracted" format layer and just be able to drop proper template files into specific directory? This way the extension could arrive to users's CiviCRM instance ready to be configured, without additional complexity of multiple extensions doing the same thing - unless I don't see some other obstacle here?

systopia:
see https://github.com/Project60/org.project60.sepa/issues/298

PR #300, opened 2015-07-14
added a proper js iban validation library

nikobo:


systopia:
will merge after v1.2 is released.

systopia:
Sorry, I'll reject this because we don't want yet another library. The best way IMHO would be to offer IBAN validation on the API and use that with AJAX.

PR #304, opened 2015-07-20
List of mandates

scardinius:
What do you think about list of mandates like that? 
In the future the list will be extended with additional columns.

Solution use `CiviCRM API3` and `AngularJS`.


tttp:
@scardinius you have mixed in that PR the file formats and list of mandates.

The best git practice is to create a specific branch for each PR, do the change there and do the PR on that branch, and checkout (or create) a new branch and checkout it if you work on another feature.

In one branch depends of the other (not the case here), it's trickier and depends of the specific case...
X+

scardinius:
Thanks @tttp for your advices about branches :-) I'll do better in next time.

Issue #313, opened 2015-08-12
Mandate List


systopia:
@scardinius recently added a draft for a mandate list (see PR-https://github.com/Project60/org.project60.sepa/pull/304) - something that has been requested multiple times.

I really like his approach and it looks good. A few remarks
* you shouldn't (have to) include ``angular.js`` yourself. Talk to @nikobo to see how it's integrated into CiviCRM - at least into versions >= 4.6
* It doesn't work yet - is that expected or did I do something wrong?
* I suggest renaming it to "Find Mandates" and add more search parameters (could be a separate ticket)
* I suggest adding a CSV export option (could be a separate ticket)

@scardinius What do you think?

scardinius:
@systopia thx for your remarks.

* 2: Results aren't displaying after opening page. You have to click on button "Search". I have to reconsider this functionality.
* 1, 3, 4: thx, sounds good

Issue #298, opened 2015-07-07
Add 'CitiBank Polecenie Zaplaty' file format


systopia:
As described in https://github.com/Project60/org.project60.sepa/pull/297, some folks want to add a new file format. However, it's obviously not done by adding the option value.

The current system has a predefined algorithm, but if you (@michau or @scardinius) can provide a smarty template for that, it wouldn't be too much work to integrate.

Also, the file format would need a sensible name, i.e. the format specification instead of the bank's name.

systopia:
@michau / @scardinius: let us know if you're interested in a cooperation.

michau:
@systopia Yes, we're interested in the cooperation. :-) Right now we're at the stage where we're learning CiviSEPA code and trying to find out what needs to be done in order to use it for CitiBank direct debit mechanism.

First of all, let's scratch the option_value SQL code in the comment - we've put it in the comment for reference, but it's practically irrelevant to broader issue. :-) Sorry for confusion. :-)

In general, seems like CitiBank in Poland has slightly different way of handling direct debit than SEPA standard (format wise), but whole process stays more or less the same. To be more specific: you submit a list of mandates for verification in specific format (format 1), receive mandate verification status info, than submit transaction requests normally in another format (format 2). Both are text with delimiters, with headers and footers.

So far, we identified following things that we would need to do in order to be able to use CiviSEPA:

1. implement ability to plug in CitiBank specific format
2. implement ability to export mandate verification file (naming convention for this file includes Client's codename given by Bank, so we'll need a format specific settings field here)
3. implement ability to parse and import both mandate verification status file and transactions status files from bank
4. potentially allow handling multiple currencies (PLN transations handled here)

We have much longer list of potential smaller changes and improvements, but leaving these aside for now to focus on base issue - adding "pluggable formats".

So - the pull request accompanying #297 (see: https://github.com/scardinius/org.project60.sepa/commit/551c56789b80ef6db88847d8a6e4f6b679adccc6) does not consist of option_value changes only - this is proof of concept refactoring trying to sketch out the concept of pluggable formats. Please take a quick look and let us know if that's something that would. Currently implemented XML format would stay as default and work as before.

We cannot fund this feature, but we'll gladly implement it and contribute the code, together with other improvements, testing effort, maybe even some significant documentation improvements.

systopia:
@michau Sorry, I was fooled by the comment - I really didn't see that the PR was much bigger :-)

[...]

PR #319, opened 2015-09-03
Generate mandates file

scardinius:
This PR includes:

* Citibank transaction templates are finished #298 
* Citibank package of mandates templates are finished
* Added new functionality: creating package of mandates. This is a additional file which is required by CitiBank. Something like FRST status of ordinary Sepa transaction file.
* Added page "Find packages"
* Added page "New packages"
    - checking whether format provides creating of packages
    - package is created based on all mandates which aren't included to other package
    - CP852 encoding
* Added possibility to change encoding of transaction file, default value is `CRM_Sepa_Logic_Format::$out_charset = 'UTF-8'`
* Moved some methods to `CRM_Sepa_Logic_Format` class
* Changed menu label `CiviSEPA Mandates` to `Find Mandates` #313


systopia:
Wow, thanks @scardinius, that's a lot of stuff! I'm going to look into this soon, but I don't think I can do it this week...

systopia:
@scardinius I'm not sure I fully understand the concept of mandate packages... Would you be open to a Skype call? Just send me an email (endres -at- systopia.de)...

scardinius:
I'd like to explain the concept here. First of all for those of us who can also have a problem with understanding meaning of packages.

Packages of mandates are similar to group of contributions (on Dashboard page).

Polish banks implement Direct Debit in a little different way than SEPA standard expected. I described below a history of one mandate in both situation:

Step | SEPA standard | Polish implementation
--- |--- | ---
1 | Mandate form is filled out by a donor | Mandate form is filled out by a donor
2 | Mandate is created in CiviSEPA | Mandate is created in CiviSEPA
3 | ... | User of CiviSEPA prepares package of mandates (the file with names, addresses and accounts)
4 | ... | The file with package is uploading to bank
5 | ... | Bank examines the mandates 
. | At first month: | At first month:
6 | User of CiviSEPA updates a recurring groups | User of CiviSEPA updates a recurring groups
7 | The file with group includes direct depit with status FIRST (steps like 3, 4 & 5 in Polish version) | The file with group include direct debit without any special status
. | At second month: | At second month:
8 | User of CiviSEPA updates a recurring groups | User of CiviSEPA updates a recurring groups
9 | The file with group includes direct depit with status RCUR | The file with group includes direct debit without any special status

Conclusion: 
* Polish implementation doesn't use field "status", uses two different files.
* Line with status FIRST in SEPA Standard has a meaning of two lines in two different files in Polish implementation.

systopia:
I see. Well, it's still close enough to include it in the module. But I would suggest to add a switch, so you would have to explicitly turn it on - otherwise it'll definitely confuse "regular" SEPA users.
@scardinius Agreed?

scardinius:
I agreed with you @systopia. Where do you want to add such switch?

First of all difference between both implementation is visible in menu - additional pages are here https://github.com/Project60/org.project60.sepa/pull/319/files#diff-5e22d46e5043ea370a50d684fd8b8fecR492

What about with a case when user has a two creditors with both formats?
I don't know a good solution now...

Proposal:
* on pages "New package" and "Find package" there is a test:
  * If you have a creditor with special format -> run normally
  * If you don't have a creditor with special format -> display a explanation.

I'm not sure whether this is a good idea.

systopia:
Ok, we"ll do the following:
 1. I created the ``dev_mandate_packages`` branch, let's develop this feature in there until it's ready to be merge. I already pulled your code in there.
 2. I created https://github.com/Project60/org.project60.sepa/issues/330 to deal with the development of this feature.
 3. I'll put a switch in this dev branch
 4. I'll close this PR

systopia:
Oh, I forgot, please create your changes as PRs for that dev branch... Or we'll just add you as a collaborator... What do you prefer?

tttp:
@scardinius is this a realistic scenario to have the two methods 'sepa' and 'pl direct debit' mixed in the same installation?

if it isn't, it might be best to keep the pl as a separate extension, that:

- allows to generate packages
- adds a hook so the the frst status is replaced by rcur

systopia:
That sounds like a good idea to me... @scardinius ?

michau:
@systopia Sorry, I've held this conversation for a while, since we didn't have clear situation with our work around here.

So coming back to the discussion: it is a realistic scenario to have two methods in the same installation. Frankly, we don't have this kind of scenario here, but with multitude of bank formats here in PL, its really possible. Right now, we have three different banks on our plate, each one with other format. Even more realistic, having local bank format (for PL based fundraising) and SEPA format (for international fundraising).

In principle, I would be very enthusiastic if we could just hide the menu item and functionality regarding mandate lists based on a setting in format. It doesn't sound like a large code complexity cost.

I'll discuss it with @scardinius tomorrow and we'll come back with further plan. If in the meantime you come up with any thoughts about the above, feel free to let us know.

systopia:
My understanding was that only the "Mandate Packages" feature would go into it's own extension, because it creates its own data structure - which 90% of users won't need.

Everything else (file formats, all your other improvements) would still be pulled into CiviSEPA. Doesn't that make sense to you?

Anyway, let me know what you guys think...

tttp:
@systopia  That was my thinking as well, sorry if I wasn't clear, but yes, we already deal with various file formats, adding a few polish ones in the mix isn't an issue

@michau Why whould you use the pl format for the national fundraising IF the bank is able to handle the SEPA one? In theory, banks should handle sepa, no matter if national or across the border. As a workaround, you have to deal with random crappy format because some banks aren't yet up to speed (it's been mandatory for only two years and in discussion for only a few decades, they clearly are caught by surprise I presume). But if they have the sepa working for international, they have to let you use it for national fundraising too, and stop making you jump through hops ;)

X+

P.S. I found the list of BICs for pl banks if you need to, and with the little bic extension, you can even hide the bic and only ask for the IBAN, so life is good.

michau:
Having mandate list as separate extension - let me look around it a bit and will let you know.

@tttp 
re: "Why whould you use the pl format for the national fundraising IF the bank is able to handle the SEPA one?"
Bank is not able. :-) We work with 4 orgs and 3 banks right now regarding direct debit and in each and every case SEPA format is out of question. "No, we don't support it, sorry. Why? Because." We have one case, where org is considering potentially getting foreign (Estonian? or Latvian? bank account) to do international DD.

re: little BIC. I found the list too, and even contributed: :-) https://github.com/Project60/org.project60.bic/pull/19

scardinius:
Hi,

I and @michau decided what to do with packages of mandates, finally. I'd like to propose such steps:

* first of all, **new extension** is already created on [SepaMandateBatch](https://bitbucket.org/caltha/pl.infozmiana.sepamandatebatch) and this is a place where all additional functionally will be moved/prepared,
* new name is `Batch of mandates`, instead of `Package`,
* current PR will be resolved:
    * #320 to close without merging,
    * #321 add die() and will be ready to merge,
* any modification will be prepared on `master` branch which means 1.3 version,
* new PR's with selective modifications

We're also interested in resolving issue #348 so I'll looking around what's going on.

What do you think?

systopia:
Sounds great!

#348 is definitely an urgent matter for us, we're trying to find funding...

scardinius:
That's all needed PRs to accomplish this task.

PR #321, opened 2015-09-04
Fixed problem with validation during deletion of contributions

scardinius:
This PR resolves problem with validation during deletion of contributions.

* better SQL query
* simple warning to user
* move code to validation hook

### DIFF
--- a/sepa.php
+++ b/sepa.php
@@ -377,27 +377,6 @@ function sepa_civicrm_merge ( $type, &$data, $mainId = NULL, $otherId = NULL, $t
   }
 }
 
-/** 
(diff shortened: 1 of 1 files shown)


systopia:
Hi @scardinius: I'm not quite happy with this. I like you approach, but it will only work when the ``delete`` is initiated via the UI. If it's called via API, it'll still produce corrupted data.

Are you o.k. with me taking over your addition, but leaving the ``die`` in? Just in case?

scardinius:
Hi @systopia,

Did you mean about such change of die()?

systopia:
Thanks @scardinius!

PR #323, opened 2015-09-07
Set up default mandate type to recurring on create form

scardinius:
SEPA Direct Debit is designed to provide recurring payments.

We found out that users often forget about changing mandate type from `One time` to `Recurring` during creating new mandate.


systopia:
@scardinius: Thanks for this. One request, though: please be more precise in your commit messages, makes it easier to understand later on...

scardinius:
Thx @systopia. Of course, I promise to be more precise in my commit messages.

systopia:
Sorry, I didn't want to be an*l about it... :)

scardinius:
No problem :-)

systopia:
I had to take the default setting back, because it would confuse customers that are already using it, in the same way it confused yours. I suggest to implement a default setting for that, see #332

PR #361, opened 2016-01-26
Add multi currency functionality

scardinius:
Multi currency functionality is required by Polish formats, for example for CitiBank. It's tested by our clients and works :-) 
This is next step of #319 

By the way, What is your prefered method of upgrading `SQL`? 

In this PR I added new column `civicrm_sdd_creditor.currency`, and only in `sql/sepa.sql` file which is used during installation. And what about upgrade?

```sql
ALTER TABLE civicrm_sdd_creditor ADD COLUMN `currency` VARCHAR(3) NOT NULL DEFAULT 'EUR' COMMENT '3 character string symbol of currency';
```

### DIFF
--- a/CRM/Admin/Form/Setting/SepaSettings.php
+++ b/CRM/Admin/Form/Setting/SepaSettings.php
@@ -116,6 +116,13 @@ class CRM_Admin_Form_Setting_SepaSettings extends CRM_Admin_Form_Setting
         // do not use array_merge() because it discards the original indizes
         $country_ids = array('' => ts('- select -')) + $filtered;
 
+        $config->defaultCurrency;
+        $currencies_enabled = CRM_Core_OptionGroup::values('currencies_enabled');
+        $currencies_dict = array('' => ts('- select -'));
+        foreach ((array)$currencies_enabled as $k => $v) {
+          $currencies_dict[$k] = $k;
+        }
+
         // look up some values
         $excld_we = CRM_Core_BAO_Setting::getItem('SEPA Direct Debit Preferences', 'exclude_weekends');
         $hide_bic = CRM_Core_BAO_Setting::getItem('SEPA Direct Debit Preferences', 'pp_hide_bic');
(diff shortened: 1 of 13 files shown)


systopia:
@scardinius sorry for the delay, I haven't forgotten you :) 
I'll probably get to your PRs in the next days...

systopia:
Two thoughts:
 1. the upgrade should use the "upgrader" class, see here https://github.com/Project60/org.project60.sepa/blob/1.2.1/sepa.civix.php#L98. The way it is now, the PR would break the system for people upgrading...
 2. multi-currency is a pretty exotic scenario for SEPA, but I can see how this cannot be a separate extension. But could we at least put this into a setting that defaults to "EUR only"? It doesn't even have to have an interface, your other extension could set it for you... I just want to avoid people getting the idea that you can collect other currencies than EUR in a regular environment, and then complain if it doesn't work :)

@scardinius: Would you be able to look into these points? 

In any case: Thanks a lot for all these contributions!

tttp:
Very valid point, as sepa "standard" is never going to do anything but
euro, worthwhile having a multi currency setting defaulting to false and no
UI in default sepa to handle that

@scardinius, could you as well in the new mysql column add a comment saying
it "requires an external extension"

scardinius:
Hi all,

Thanks a lot for review. I agree with you. 

* I think that by using such code I can check external extension and not displaying currency at all, look at here
[SepaSettings.php](https://github.com/scardinius/org.project60.sepa/commit/a2915433acf8e7e4ea1bf464f68e5f2c7f817a66#diff-e0f73f62116f995802611723bbabc397R119)

```php
$query = "SELECT count(id) AS test
          FROM civicrm_extension
          WHERE full_name = 'pl.infozmiana.sepamandatebatch' AND is_active = 1";
if (CRM_Core_DAO::singleValueQuery($query) == 1) {
  // add additional dictionary and field 'currency' on sepa settings page
}
```

* Yes, I'll add `required by external extension` to comment of currency column
* and the last `Upgrader.php` class. I see that this class doesn't use in Sepa now. It's not problem to add this class with simple sql alter table code. But I'm not sure how this code will be triggered. Do you have a example of code?

scardinius:
Ok, I tested upgrading on my own simple extension. I've already know everything what is needed to prepare such Upgrader class. I'll do new PR for this.

scardinius:
Ok, all done. What do you think about such changes?

systopia:
> Ok, all done. What do you think about such changes?

It sounds great, but I won't have the time to look into this until Tuesday. Sorry. But I'm looking forward to it, I'll let you know! Thanks again!

systopia:

PR #362, opened 2016-01-26
Improve list Find Mandates

scardinius:
This is a simple improvement of list **Find Mandates** according to #319.


systopia:
@scardinius Thanks again!

PR #363, opened 2016-01-26
Fix filename when download again after close group

scardinius:
Extension of file is configured on level of creditor and therefore in column `civicrm_sdd_file.filename` is a filename with proper extension. For example:

* `CITIBANK-TXG-2-FRST-2016-01-26`**.txt** in Citibank format
* `SEPA-TXG-2-FRST-2016-01-26`**.xml** in standard Sepa format

#319


systopia:
@scardinius thx

PR #364, opened 2016-01-26
Prepare generating file of group for new formats

scardinius:
The commit [eea713a](https://github.com/Project60/org.project60.sepa/commit/eea713a50a898cd2babbbd0af7c887f16a69e771) includes preparations for new formats like CitiBank. Current standard SEPA format works as well.

This one of the last PR according to #319. Remain to do two formats, each of them in separate commits/pr.


systopia:
@scardinius: I'm assuming this needs PR #361, correct? I'll wait for that to be resolved.

PR #369, opened 2016-02-04
Add upgrader class for new column currency

scardinius:
Required by #361 
[2914dc7](https://github.com/Project60/org.project60.sepa/commit/2914dc74132b503d7fa93f222b2b90e1b2ec9445)


systopia:

PR #370, opened 2016-02-05
Add checking if external extension is enabled for currency

scardinius:
Required by #361 
commit [647cce4](https://github.com/Project60/org.project60.sepa/commit/647cce4f7cab9aab0719851046aab7cd821aca3c)

External extension is [Sepa Mandate Batch](https://bitbucket.org/caltha/pl.infozmiana.sepamandatebatch)

If this extension is installed, then Setting page has additional field: Currency


systopia:

Issue #371, opened 2016-02-09
Add multi currency functionality


systopia:
@scardinius wants to add multi-currency functionality for a Polish scenario.The development is currently in https://github.com/Project60/org.project60.sepa/tree/dev_361, see also:
- PR https://github.com/Project60/org.project60.sepa/pull/361
- PR https://github.com/Project60/org.project60.sepa/pull/369
- PR https://github.com/Project60/org.project60.sepa/pull/370

systopia:
@scardinius: I pulled your PRs into https://github.com/Project60/org.project60.sepa/tree/dev_361 and resolved some minor conflicts. I still have three more items I'd like to discuss:
 1. Why did you chose ``0010`` as the upgrader revision? Wouldn't it make sense to tie that to the extension version, i.e. ``0130`` for 1.3.0?
 2. When I said I wanted a setting, I was rather thinking about a checkbox on the settings page instead of hardcoding your extension ID into it. I can implement that if it's fine with you..

scardinius:
Hi @systopia

1. I agree `0130` is better value for revision.
2. I misunderstood you and @tttp. I hoped It will be better when I hide currency option at all.

scardinius:
Hi @systopia 

I made PRs with required changes.

1. Change upgrader revision to 1.3.0
2. Rebuild checking if user needs multi currency mode.
    * Prior: hardcoded checking of external extension
    * Current: additional parameter `Multi currency mode (not only EUR)`, default false.

I hope This will be fine with you :-)

systopia:
> I hope This will be fine with you :-)

Thank you again! I merged it, but I'll leave the ticket open until we tested it successfully :)

scardinius:
Hi @systopia 

From my side, this feature works fine.

systopia:
Ok, great!

systopia:
Needs to be merged again into ``1.4``

systopia:
I've decided to go for a new implementation, see ``dev_371`` branch: one currency per creditor, no mixed currency within the same creditor allowed.

Currently being tested.

PR #375, opened 2016-02-18
Fix upgrader revision to 1.3.0

scardinius:
Fix upgrader revision to 1.3.0 #371
and rebuild checking if user needs multi currency mode.

Prior version: hardcoded checking of external extension
Current version: additional parameter `Multi currency mode (not only EUR)`, default false.


systopia:
@scardinius Great! I'll do some testing in the dev branch (probably not this week, though) and then merge the whole thing into ``master``, ok?

scardinius:
@systopia ok.

PR #378, opened 2016-03-01
Add legend and change colors on dashboard

scardinius:
Hi @systopia,

We had problem with understanding meaning of colours, especially without any legend. So we decided to create own colours based on logic in code. First of all In this commit colours were moved from background to font. Second, on bottom of page is presented legend.

In examples below there aren't all cases.

What do you think?

**Before active:**
sepa-before-active

**Before closed**
sepa-before-closed

**After active**
sepa-after-active

**After closed**
sepa-after-closed


systopia:
That looks pretty cool! Thanks!

PR #379, opened 2016-03-01
Add refreshing contributions tab after creating new mandate

scardinius:
This is additional UI effect. User can see that mandate is already created.


systopia:
Great!

Issue #383, opened 2016-03-14
closing a OOFF group might end up closing unrelated mandates


tttp:
So interesting bug: we had a RCUR contribution that was never debited, and it seems that the mandate for RCUR was closed when we closed an (unrelated) OOFF contribution that had the same ID than the recurring contribution

@scardinius analysis
https://github.com/Project60/org.project60.sepa/blob/master/CRM/Sepa/Logic/Group.php#L56
we close all mandates, not only OOFF mandates

scardinius:
My proposal in order to resolve problem is here #386

PR #386, opened 2016-03-21
Fix problem with closing unrelated mandates #383

scardinius:
If you can find mandate with such combination of type and status, this means you have a broken mandates:

type | status (invalid) | change to
--- | --- | ---
OOFF | RCUR | OOFF
RCUR | SENT | FRST or RCUR

* `RCUR` status is reserved only for recurring transactions
* `SENT` status is reserved only for one-off transactions

Be careful during modifying your data in db. Your cases might be slightly different.


systopia:
Looks sensible... thanks

Issue #351, opened 2015-12-16
Compatibility with civicrm 4.7


magnolia61:
Hi Bjorn, I'm posting this issue as a placeholder for civicrm 4.7 compatibility. From what is within my possibilities I want to pro-actively try to help get civisepa to work with 4.7 (which is almost stable, now at beta3). In this issue we can try to collect the opportunities  and pitfalls for civisepa with civicrm 4.7

magnolia61:
When opening the admin settings (civicrm/admin/setting/sepa?reset=1) I get the following fatal error in 4.7:
Fatal error: Call to undefined method CRM_Core_Config::countryLimit() in /home/ozklab/domains/onvergetelijk.nl/public_html/porto/sites/all/modules/civicrm_extensions/org.project60.sepa/CRM/Admin/Form/Setting/SepaSettings.php on line 101

nikobo:
#352 is part of this, at least for beta 2

systopia:
Hi @magnolia61. Thanks for your initiative. We haven't started looking into this, but afaik they fundamentally changed the payment processor structures (again), so I expect some problems there.

magnolia61:
Some observations running in civicrm 4.7:
- When I try to go to the civisepa setting I get:
   Fatal error: Call to undefined method CRM_Core_Config::countryLimit() 
- Updating a batch in the civisepa dashboard seems to go ok, but updating it a 2nd time gets:
  Notice: Undefined index: id in civicrm_extensions/org.project60.sepa/CRM/Sepa/Logic/Batching.php on line 433

magnolia61:
I know 4.7 compatibility is still not available/funded, so purely FYI and to document for future use:
When (in civicrm 4.7.7) I try to look at the contributions of an active group I get
Cannot read SEPA transaction group [350]. Error was: 'A fatal error was triggered: is not of type String'
The url is: https://www.onvergetelijk.nl/civicrm/sepa/listgroup?group_id=350

systopia:
@magnolia61: Yes, we still don't have anyone willing to fund this. But thanks for documenting...

PR #388, opened 2016-03-29
Fix undefined countryLimit() in CiviCRM 4.7 #351

scardinius:
Method `CRM_Core_Config::countryLimit()` was moved to `CRM_Core_BAO_Country::countryLimit()` in version 4.7. This fix checks If we are working on 4.7 or other version and use proper code.

Explanation:
https://github.com/civicrm/civicrm-core/commit/0acb7f15552889bdba24b013da4dafebebf3a838
https://github.com/civicrm/civicrm-core/commit/90777b73c3a04e023c36417abf3b4ab7a282dfb9


systopia:
Great, thanks! Could you please use ``version_compare()`` though? See e.g. https://github.com/Project60/org.project60.sepa/blob/1.2.2.beta1/CRM/Utils/SepaCustomisationHooks.php#L44  ...although your approach is pretty ingenious :)

scardinius:
You're right, code with `version_compare()` looks better :-)

systopia:
> You're right, code with ``version_compare()`` looks better :-)

It does, doesn't it? :-)

Fabian-SYSTOPIA:
I can confirm that the fix is working for (CiviCRM 4.7.12)

Issue #414, opened 2016-12-16
Settings for payment processors are not unique in 4.7


scardinius:
CiviCRM 4.7 doesn't use `civicrm_setting.group_name` in order to distinguish settings. This means that `civicrm_setting.name` has to be unique! 

Sepa has settings `pp1`, `pp2`... as a container for creditor_id for selected payment processor. There were also different group_name, 'SEPA Direct Debit PP' and 'SEPA Direct Debit PP Test'.

After upgrade below code returns the same value (based on a greater civicrm_setting.id)

```php
$creditor_id = CRM_Core_BAO_Setting::getItem('SEPA Direct Debit PP', 'pp'.$pp_id);
$test_creditor_id = CRM_Core_BAO_Setting::getItem('SEPA Direct Debit PP Test', 'pp'.$pp_id);
```

I'd like to change name into

* `pp_live1`
* `pp_test1`

systopia:
Thanks @scardinius for reporting. Would you mind creating a PR?

scardinius:
@systopia yes, I'm working on it now :)

scardinius:
@systopia This fix handles getting and setting values but not migrating from old names to new. Do you have good idea? After migrating to 4.7 it's not possible for script/php to decide which value is for live or test payment processor...

systopia:
> Do you have good idea? After migrating to 4.7 it's not possible for script/php to decide which value is for live or test payment processor...

I know, we had the same problem in another extension (de.systopia.donrec). We ended up putting upgrader code into the extension and added the instruction to upgrade the extension (on 4.6) _before_ upgrading CiviCRM (to 4.7)...

systopia:
released in 1.2.4

PR #415, opened 2016-12-16
Fix settings for payment processors in 4.7

scardinius:
#414


systopia:
Thanks @scardinius! I used a slightly abridged version of your fix (https://github.com/Project60/org.project60.sepa/commit/32fd2b8195990a88675c948e49dadbf000560c9b) so that it doesn't break running payment pages.

Those fixes will be merged into ``master`` as soon as 1.2.4 is released.

Issue #411, opened 2016-11-29
PHP7 compatibility: Error: Call to undefined function split() 


magnolia61:
Thanks for the 1.2.4 beta1 release. Just tried it on drupal 7.5.2 civicrm 4.7.13 and PHP7
Using the civisepa admin settings I get the following error when saving the default values:

Error: Call to undefined function split() in CRM_Sepa_Logic_Settings::getListSetting() (line 86 of /home/xyz/domains/ourdomain.nl/public_html/sites/all/modules/civicrm_extensions/org.project60.sepa/CRM/Sepa/Logic/Settings.php).

Somewhere I found that for PHP7 the split function is deprecated and explode can be used instead.
Just reporting FYI ;-)

tttp:
Hi,

Thanks for the report. Is the rest of civi working in PHP7? 
the latest update I have was that it relies on several key libraries (eg pear ones) that aren't compatible with PHP7

magnolia61:
Working flawlessly in production for a month now, using even registrations and contributions. And: all much faster!

systopia:
@magnolia61: Thanks for the report. Care to share you patch?

magnolia61:
I have no patch, just reporting.

DanielvV:
Works for me

systopia:
Released with 1.2.5.beta1

PR #436, opened 2017-07-04
Make it compatible with php7

DanielvV:
Replace deprecated (php5) / removed (php7) function split() with explode() or preg_split()


systopia:
I think this has been fixed with https://github.com/Project60/org.project60.sepa/issues/411, and is currently in 1.2.5.alpha1.

As stated in the https://github.com/Project60/org.project60.sepa/blob/master/README.md the master branch is not suitable for deployment. We would like to change it to the latest release but some people are using it, so it might cause trouble with them since it would be a step back. But I guess we'll to it soon anyway, apparently a lot of people don't read the readme.

DanielvV:
Thanks, looked right over 1.2.5.alpha1 on the releases page.
Sorry that i didn't search before submitting a patch. It's exactly the same as #411 .

systopia:
No problem, thanks for participating anyway :)

PR #453, opened 2017-10-03
Small fix for `ts()` call

DanielvV:


systopia:
Well spotted, @DanielvV, thanks. And a slightly embarrassing copy/paste error  :)

PR #454, opened 2017-10-04
Replace deprecated `CRM_Core_OptionGroup::getValue()`

DanielvV:
with `CRM_Core_PseudoConstant::getKey()`
or  `CRM_Core_PseudoConstant::getName()`
or  `CRM_Core_PseudoConstant::getLabel()`


DanielvV:
I am not so sure about [CRM/Sepa/Page/CreateMandate.php:250](https://github.com/Project60/org.project60.sepa/pull/454/files#diff-9f85cd957b2156394bfa30a06ebc5237R250)

DanielvV:
Fixes #455

systopia:
Cool, thanks @DanielvV. Did you verify that it works on CiviCRM 4.6 and 4.7?

DanielvV:
@systopia Only on 4.7

DanielvV:
I am not sure if casting as `(int)` is needed with [`getKey()`](https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/PseudoConstant.php#L463)

systopia:
> I am not sure if casting as (int) is needed with getKey()

I usually did that when I wanted to use it in an SQL query later. (that was before I discovered CiviCRM's parameter types).

@DanielvV Which location are you referring to?

DanielvV:
@systopia In almost every instance of the use of getValue() (now getKey())

PR #467, opened 2018-02-06
Don't compute tokens if not used

### LINKED ISSUE #466: Token values computed even if they are not use
The value for the tokens provided by the extension is computed for each contact of the mailing, even if the mailing does not use any of the tokens. This has 2 drawbacks:

 - Performance overhead when the mailing is sent. This is normally small, unless the list of SEPA donors is really large.
 - More importantly: if there is a fatal error while computing the token values, this prevents the mailing from being sent even though it shouldn't matter.

For example, we had some inconsistent data (some mandates pointing to deleted contribution), and because of that we couldn't send any mailing, even mailing where all recipients had clean data, because the failing mailing was blocking all others in the queue. In that sense, I think this is a serious issue.

Since I wrote a fix to unblock our mailings, I will submit a patch.

rthouvenin:
Fix #466

### DIFF
--- a/sepa.php
+++ b/sepa.php
@@ -695,6 +695,9 @@ function sepa_civicrm_tokenValues(&$values, $cids, $job = null, $tokens = array(
   try { // make sure this doesn't cause any troubles
     $prefix = ts("Most Recent SEPA Mandate", array('domain' => 'org.project60.sepa'));
     $prefix = str_replace(' ', '_', $prefix); // spaces break newletters, see https://github.com/Project60/org.project60.sepa/issues/419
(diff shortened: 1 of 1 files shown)


tttp:
@rthouvenin, I'm with @systopia , could you try testing by setting up and sending one mailing that has a sepa token, to check that it's only skipping the token when they are not used?

rthouvenin:
I did test it, the token was rendered as expected

systopia:
> I did test it, the token was rendered as expected

That's good enough for me :) Thanks!

PR #475, opened 2018-03-06
Fix #472 invalid SQL query

### LINKED ISSUE #472: Invalid SQL query in prepareCreateForm(): for SELECT DISTINCT, ORDER BY expressions must appear in select list
The MySQL query defined in CRM/Sepa/Page/CreateMandate.php:239 is invalid for MySQL with ONLY_FULL_GROUP_BY, and for other SQL dialects as well:

```sql
SELECT DISTINCT iban, bic FROM civicrm_sdd_mandate WHERE contact_id=$contact_id ORDER BY creation_date DESC;
```
This causes the mandate form (e.g. for adding a mandate for a contact) not opening.

Including the ORDER BY field `creation_date` in the SELECT clause appears to solve the problem without side effects:

```sql
SELECT DISTINCT iban, bic, creation_date FROM civicrm_sdd_mandate WHERE contact_id=$contact_id ORDER BY creation_date DESC;
```

rthouvenin:
Rather than adding the order field to the select clause as suggested in the issue, I simply removed the order clause since it is not used nor any other field from the mandate.

I also parameterised the query, as the contact id comes from a GET parameters in some scenarios, so it is prone to SQL injection.

### DIFF
--- a/CRM/Sepa/Page/CreateMandate.php
+++ b/CRM/Sepa/Page/CreateMandate.php
@@ -236,8 +236,9 @@ class CRM_Sepa_Page_CreateMandate extends CRM_Core_Page {
 
     // look up account in other SEPA mandates
     $known_accounts = array();
-    $query_sql = "SELECT DISTINCT iban, bic FROM civicrm_sdd_mandate WHERE contact_id=$contact_id ORDER BY creation_date DESC;";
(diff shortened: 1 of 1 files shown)


scardinius:
indeed, better than #474

systopia:
Thanks guys.

PR #476, opened 2018-03-22
Fix cancelling mandates by using statuses Pending and In Progress

scardinius:
Before: SEPA searches for only Pending contributions

After: SEPA searches for Pending or In Progress contributions


bjendres:
Thanks @scardinius, but this is not right. In the current SEPA logic, status ``In Progress`` means, that the contribution has already been sent to the bank. If that's the case, we cannot simply delete it.

I know the current CiviCRM version interferes with that, see e.g. #514, but we shoudn't be deleting contributions that have been sent to the bank.

What do you think?

Issue #479, opened 2018-04-04
SDD payment processor doesn't use Exclude Weekends setting


scardinius:
Hi,

When cycle day is in weekend then collection date is deferred to the nearest Monday date (with assumption that `Exclude Weekends` setting is on). This behavior works fine during updating recurring mandates from sepa dashboard. 

However it doesn't work when mandate derived from contribution page configured with sepa payment processor because first contribution is created with undeferred date

Example: cycle day 21, collection date 2018-04-21 is Saturday, therefore it should be defered to 2018-04-23. 

In result mandate has two contributions:

* created during filling out a contribution page with date 2018-04-21 which is not included in sepa group
* created during updating recurring mandates with date 2018-04-23 which is included in sepa group

systopia:
Thanks, @scardinius, I'll release it with @1.2.6@

systopia:
Should be released with ``1.4``

PR #480, opened 2018-04-04
Use Exclude Weekends setting in SDD payment processor

scardinius:
#479


systopia:
Great, thanks @scardinius

PR #504, opened 2018-09-19
Small fix php

adanielvv:
Is needed to suppress `Warning: Use of undefined constant yes - assumed 'yes' (this will throw an Error in a future version of PHP)`


adanielvv:
Yes, that is better. I just tested it.
`{if $can_delete eq yes}` is equivalent to:
- `{if $can_delete eq "yes"}`
- `{if $can_delete eq "thisissomething"}`
- `{if $can_delete}`

systopia:
Great, thanks @adanielvv

PR #531, opened 2019-04-23
Fix certain retry date range options not being applied

pfigel:
Some of the date range options used for retries include the current time in their respective value. Because time does not stop between the moment the form is loaded and when it's submitted, the options generated during submission will have different values, meaning QuickForm will discard the submitted form value as it's not a valid option for the field.

This results in no date filter being applied.

This is fixed by setting the time to midnight for the affected options, which arguably makes more semantic sense anyway.


bjendres:
Thanks @pfigel

PR #532, opened 2019-04-26
Change PSP BIC handling in form and during validation

pfigel:
With https://github.com/bjendres/de.systopia.pspsepa/pull/6, `de.systopia.pspsepa` uses the BIC field to store (merchant) account names.

To support this in CiviSEPA, this changes the mandate form to stop hiding the BIC field for PSP creditors and renames the field instead. This also alters the BIC format check for PSP creditors to allow values up to 25 characters.


bjendres:
Thanks @pfigel

Issue #535, opened 2019-06-10
Link CiviSEPA Dashboard should have unique navigation name


scardinius:
We have situation that two links in menu have the same navigation name

Contributions > Dashboard
Contributions > CiviSEPA Dashboard

So it's not possible to add submenu to CiviSEPA Dashboard by external extension.

PR #536, opened 2019-06-10
Fix Dashboard menu to have unique name

scardinius:
related to #535


bjendres:
Thanks @scardinius, i can't believe nobody noticed that so far...

PR #539, opened 2019-08-14
Fix recurring contribution link removal for mandates

pfigel:
This fixes an issue with the recurring contribution action link removal in versions of CiviCRM where recurring contributions are split across two tables (active/inactive).


bjendres:
Thanks @pfigel

bjendres:
Will be released with ``1.5``

PR #542, opened 2019-08-28
Fix inconsistent formatting and validation of IBAN/BIC

pfigel:
This fixes some inconsistencies in the way IBANs and BICs are formatted and validated in different components. The main issue was that IBANs for PSP creditors would be formatted according to SEPA/IBAN logic (i.e. uppercase, remove spaces), which causes issues for case-sensitive PSPs. Instead, this is now only applied for SEPA creditors.

The IBAN/BIC formatting was moved to a central function to DRY up the code. The formatting is now also applied consistently e.g. in the new mandate form, where previously a lower-case IBAN would've been rejected as invalid.

This also contains some whitespace cleanup that my IDE seems to insist on.


bjendres:
Thanks @pfigel

bjendres:
Will be released with ``1.5``

PR #544, opened 2019-09-10
Fix PSP token format validation being too strict

pfigel:
PSP tokens have practically no format requirements, so we now accept anything that's a string and not empty.

Reference: GP-3505


bjendres:
Will be released with ``1.5``

Issue #549, opened 2019-10-29
Create PAIN.008.001.02 format without (real) BICs


bjendres:
@pfigel asked for the following adjustment in ``pain.008.001.02``:

Instead of
```
        <DbtrAgt>
          <FinInstnId>
            <BIC>NOTPROVIDED</BIC>
          </FinInstnId>
        </DbtrAgt>
```
we should be using:
```
        <DbtrAgt>
          <FinInstnId>
            <Other>
              <Identification>NOTPROVIDED</Identification>
            </Other>
          </FinInstnId>
        </DbtrAgt>
```

bjendres:
Will be released with ``1.5``

PR #550, opened 2019-10-29
Change BIC-less pain.008.001.02 format to never set BIC

pfigel:
This changes the alternative pain.008.001.02 format to always set the "no bic" value to NOTPROVIDED, rather than using the stored BIC.

To make the intent more clear, this also renames the format to "pain.008.001.02 without BIC".

### DIFF
--- a/resources/formats_option_group.json
+++ b/resources/formats_option_group.json
@@ -104,7 +104,7 @@
     {
       "_lookup": ["name"],
       "_translate": ["label"],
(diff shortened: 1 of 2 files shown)


bjendres:
changing objective of #549

bjendres:
Droppend in favour of https://github.com/Project60/org.project60.sepa/commit/4d9589672951b6110cb61c68802b1a597db9c0a4

PR #551, opened 2019-11-07
Fix wrong tags being used in BIC-less pain.008.001.02

pfigel:
This replaces `<Other>` and `<Identification>` with the correct tags `<Othr>` and `<Id>`.


bjendres:
fix for #549

Issue #555, opened 2019-11-28
Add hook for calculating/altering the next calculation date of recurring contribution


jaapjansma:
My use case is the following: 

I have a membership with a fixed period from January to December.
When a new member signs up in October I want to collect the money in October. However the next contribution should be collected in January. As all membership fees are collected in January.

At the moment I can use Yearly as an interval but then the contribution for the membership is always collected in October.

bjendres:
If you're quick with a PR, you'll make it into ``1.5`` :)

jaapjansma:
I have almost finished the PR. Just needs testing. Hopefully that is done by the end of today or monday.

jaapjansma:
I have called the hook: `hook_civicrm_alter_next_collection_date` with two parameters the `$next_collection_date` and `$mandate_id`

What do you think of the name of the hook?

You can see my changes here: https://github.com/Project60/org.project60.sepa/compare/master...jaapjansma:dev_555

bjendres:
> What do you think of the name of the hook?

good :)

> I have called the hook: `hook_civicrm_alter_next_collection_date` with two parameters the `$next_collection_date` and `$mandate_id`

Also ok, but I wonder if we can pass more data (e.g. the loaded mandate), otherwise we might seriously slow down the batching process, if the hook has to load thousands of mandates individually to find out whether to do anything about it...

jaapjansma:
I agree but I discovered that contribution_recur_id which I needed (because that one was stored at the membership) is not included

jaapjansma:
Oh I am mistaken, it is called mandate_entity_id.

jaapjansma:
I have updated the PR. So we pass the whole data array and I have added creditor_id

eth4ever:
Maybe it could be useful to be able to alter the amount of the first collection as wel (or the amount of the second collection). Some of our clients use a different amount when a membership start i.e. in October and the membership fee for the next year will be collected in January.

jaapjansma:
@eth4ever that is already possible with the current hooks such as the pre hook. And with the sepa installment create hook.

jaapjansma:
I discovered a mistake in my PR earlier. I will submit a PR which will fix this.

bjendres:
Will be released with ``1.5``

PR #558, opened 2019-12-10
Fixed next execution time

jaapjansma:
Another fix for #555


bjendres:
Will be released with ``1.5``

Issue #559, opened 2019-12-17
CiviSepa broken after update to 5.20.0 with localized decimal settings


jaapjansma:
**How to reproduce**

1. Install CiviCRM **version** 5.20.0
2. Set under localisation --> Languages, Currency, Localisation --> Decimal Separator to ',' 
2. Install latest version Civi Sepa
3. Create a mandate which should be collected quite soon
4. Press Update Recurring in the CIviSepa Dashboard

**Actual Results**
No prenindg contribution are created

**Expected result**
One ore more pending contributions are created

**Cause of the problem**
The `contribution.create` api fails (at https://github.com/Project60/org.project60.sepa/blob/master/CRM/Sepa/Logic/Batching.php#L187)
with the following result:

```json
    {


    "is_error": 1,
    "error_message": "total_amount is  not a valid amount: 250.00"
}

```

jaapjansma:
Ideally this should be fixed in core but an easy fix is possible by converting the amount with `(float)` before calling the api.

jaapjansma:
This has nothing to do with civicrm 5.20 but with the civicrm settings for decimal and thousand separtor. If you enter , as a decimal separtor and [space] as a thousand separator then this error will occur.

jaapjansma:
See as well: https://lab.civicrm.org/dev/core/issues/1522

bjendres:
Hm, this is going to be tricky. I remember there being a couple of changes in the way CiviCRM parses monetary strings, and CiviSEPA needs to be able to deal with all the different ways in CiviCRM ``5.x``.

Until now, passing the string to the API as a "computer formatted" number (decimal point, no thousand separator) has worked. If this is broken, then we have a major problem.

jaapjansma:
The real issue is in core and using a weird decimal separator.

bjendres:
Hm, close this ticket then?

jaapjansma:
Yes will do

PR #560, opened 2019-12-17
#559: issue with localized version of civicrm 5.20

jaapjansma:
See issue https://github.com/Project60/org.project60.sepa/issues/559


jaapjansma:
The real issue is in core. So will close this.

PR #569, opened 2020-04-02
Fix call to undefined function strotime

pfigel:
This fixes a typo (`strotime` instead of `strtotime`).

Note: I didn't do any testing on this, just noticed this when it broke a unit tests in de.systopia.contract.

### DIFF
--- a/CRM/Sepa/Logic/Batching.php
+++ b/CRM/Sepa/Logic/Batching.php
@@ -589,7 +589,7 @@ class CRM_Sepa_Logic_Batching {
       return NULL;
     }
     // ..or the cancel_date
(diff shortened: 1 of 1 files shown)


bjendres:
@pfigel thanks, that's pretty embarrassing. Just checking: you wouldn't happen to have the time/resources/passion to add a little unit test...?

pfigel:
@bjendres I'll take rain check on that 😅but I can put it on my list for the next time I change something in CiviSEPA.

bjendres:
@pfigel you're excused, especially since you guys have funded *all* of the automated testing in CiviSEPA so far...

Issue #572, opened 2020-05-27
PSP Creditors should have different payment instruments


bjendres:
The PSP (payment service provider) creditor type as introduced in #485 should also be able to have different payment instruments that the default FRST/RCUR/OOFF ones.

Implementation:
1. Every creditor gets two now settings: 
1.1 ``pi_ooff`` list of payment instruments for OOFF collections
1.2 ``pi_rcur`` list of payment instruments for RCUR collections. Here, an entry could also be a tuple, e.g. ``6-7``, representing the FRST and RCUR payment instrument
2. The default would be the current behaviour
3. Is ``pi_ooff`` or ``pi_rcur`` empty, no new mandates can be created of the respective type
4. If ``pi_ooff`` or ``pi_rcur`` has multiple values, the payment instrument has to be specified upon creation (in the form as well as the API)

PR #577, opened 2020-06-15
Find existing contributions when the payment instrument of the mandate differs

jaapjansma:
**Description**

I have a use case where we use CiviSepa for Vipps Recurring (a norwegian payment system). 
We have created mandates with the payment instrument Vipps Recur instead of Sepa DD Recurring Transaction.
This is causing two kind of troubles:
* Contributions are created multiple times because the existing contribution could not be found because those are not paid with Vipps Recur 
* Or they are removed from existing groups. 

**Previous**

The Sepa code was looking for existing contribution for a mandate based on the payment instrument Sepa DD Recurring Transaction

**After**

* Looking up existing contributions is done based on the payment instrument of the mandate.
* I also have added a possibility to set a different payment instrument in the `CiviSepaMandate.Createfull` api
* And the contributions created by Sepa will get the payment instrument from the mandate

**Comment**
When the status is FRST we stil use the `Sepa DD First Transaction` as a payment instrument.

### DIFF
--- a/CRM/Sepa/Logic/Batching.php
+++ b/CRM/Sepa/Logic/Batching.php
@@ -184,7 +184,7 @@ class CRM_Sepa_Logic_Batching {
               "contribution_status_id"              => $mandate['rc_contribution_status_id'],
               "campaign_id"                         => $mandate['rc_campaign_id'],
               "is_test"                             => $mandate['rc_is_test'],
-              "payment_instrument_id"               => $payment_instrument_id,
+              "payment_instrument_id"               => $mandate['rc_payment_instrument_id'],
             );
           $contribution = civicrm_api('Contribution', 'create', $contribution_data);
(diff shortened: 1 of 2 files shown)


bjendres:
Thanks. This issue should be covered by #572, I'll see if it makes sense to still include this particular PR.

jensschuppe:
@jaapjansma Is this still relevant with #572 (PSP support) being released with `1.6`?

bjendres:
@jaapjansma is this still relevant?

jaapjansma:
I do think so. But it is too long ago to decide on this. So feel free to close this PR or merge it.

bjendres:
> I do think so. But it is too long ago to decide on this. So feel free to close this PR or merge it.

Thanks. I think I'll close it for the time being. If somebody needs it after all, they would have to invest some time to bring it up to code.

Issue #566, opened 2020-02-19
Store IBAN in CiviBanking account


bjendres:
There should be (if CiviBanking is enabled) an option to pick a preferred behaviour:
* don't store
* store upon creation
* store upon (first) completion

Issue #96, opened 2014-01-02
Different billing name and address


Detsieber:
For credit card payments, it is possible within CiviCRM to enter a billing name and address, that differs from the name of the contact.

This should be possible for SEPA DD paymnts as well.

PR #581, opened 2020-07-11
Add separate account holder field for mandates

blipp:
Pull request by @kadl and me.

Partly addresses #96 by adding a separate account holder to SEPA mandates. The field is optional. For SEPA XML generation, the account holder is used as display name if it is not empty. If it is empty, the contact's is used as before.

Added test case for mandate creation with non-empty account holder. We did not run the tests because we do not have a working test environment set up (and you mention in the repository that you use a homegrown internal test system).

Manually tested:
- creating a mandate with (and without account holder) via the web UI
- display a mandate with (and without account holder) via the web UI
- update/replace a mandate with (and without account holder) via the web UI
- creating a mandate with (and without account holder) via create and createfull APIs
- generate SEPA XML with (and without account holder) via web UI

Edited to add:
- We did not adapt the “known bank account” feature to fill out account holders of previously used bank accounts. The reason is that CiviBanking currently does not have an account holder field for bank accounts. Possibly related issue for this: #566.
- We did not add a test case for SEPA XML generation because there are no tests for this yet, in general.


bjendres:
Thanks @blipp  and @kadl, that's pretty cool! 

I'll have a closer look and get back to you with some feedback, before merging this into ``master``. We also might have to rebase when other ongoing developments (like #572) are completed.

bjendres:
Hi @kandl and @blipp. Thanks for the PR, that looks pretty good... 

However, in light of the fact that we have to release ``1.6`` asap because of #616, I'll push this to version ``1.7`` - it seems clean to me, but I'd like to risk of compatibility issues as small as possible. Plus, there's already some stuff in there that has the potential for this.

I suggest to create a ``1.7-alpha`` with this PR for you as soon as ``1.6`` is released, ok? Feel free to ping be again, should I forget :)

bjendres:
I have now (after some unrelated issues with ``1.6``) merged it to ``master``, please let me know if this works for you.

I have two additional questions, though:
1. Could you provide some unit tests?
2. Can you think of a reasonable upgrade path, e.g. filling the account holder with the contact's display name?

jensschuppe:
@bjendres Should this be further worked on in this merged PR or in a follow-up? Since this got merged, I'd advocate for setting it to `status:fixed`.

jensschuppe:
@bjendres  Also, there are three related branches: `issue/581`, `issue/581_1.7`, and `issue/581_raw` that all got merged eventually - can we delete them?

Issue #586, opened 2020-08-05
Not possible to change the cycle day of a mandate


jaapjansma:
Currently it is not possible to change the cycle day of a mandate.

bjendres:
released with ``1.8.0``

PR #587, opened 2020-08-05
Issue #586: possible to edit the cycle day of an mandate

jaapjansma:
**Before**

Not possible to change the cycle day of an existing mandate

**After**

When Modify of Mandate is enabled and you edit an existing mandate you can also change the cycle day.

See #586

### DIFF
--- a/CRM/Sepa/BAO/SEPAMandate.php
+++ b/CRM/Sepa/BAO/SEPAMandate.php
@@ -473,6 +473,7 @@ class CRM_Sepa_BAO_SEPAMandate extends CRM_Sepa_DAO_SEPAMandate {
    *  - amount
    *  - campaign_id
    *  - financial type
+   *  - cycle_day
    *
    * Changes will take effect out to all future contributions,
    *  including already created ones in status 'Pending'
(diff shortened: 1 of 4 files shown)


bjendres:
I'm wondering whether we should add a warning, that this could substantially meddle with the collection cycles - that was the reason we didn't allow that in the first place.

jaapjansma:
You can only change the cycle day when you have enabled modifications of mandates in CiviSepa Settings. Do you still want a little warning? Or is that enough caution?

bjendres:
> You can only change the cycle day when you have enabled modifications of mandates in CiviSepa Settings. Do you still want a little warning? Or is that enough caution?

I think I'd put some strong-worded warning in this option's help text. If you're up for it, please add it. If not, I'll do it myself at some point. In any case we'd have to wait with the merge until after CiviSEPA 1.5 is released.

jaapjansma:
I have added an help icon with help text. Do you think this is enough warning?

bjendres:
> I have added an help icon with help text. Do you think this is enough warning?

Yes, thanks :)

I might also adjust it to something like: "When you change the cycle day be aware that you might end up collecting twice in the same month, or miss a collection altogether". What do you think?

In any case, I hope I can close the lid on ``1.5`` soon, and merge this...

jaapjansma:
Yes that is a bit more clear. I have updated the help text.

jensschuppe:
This will be merged for `1.8.0`.

PR #588, opened 2020-08-05
improved of text elements in the edit mandate screen

jaapjansma:
Added `crm-form-text` class on text elements on the edit mandate screen. So that it becomes a bit more user friendly.

**Before**

Screenshot_20200805_104740

**After**

Screenshot_20200805_104718


bjendres:
Thanks @jaapjansma!

PR #591, opened 2020-09-04
GP-4897 Fix array to string conversion for default PIs

pfigel:
This fixes an array to string conversion issue when default PIs are used. There's some code in `DB_DataObject` that uses these values as strings. Likely not an issue from the UI as it always passes these values as strings.

Noticed this because it was breaking tests in `de.systopia.contract` due to `convertNoticesToExceptions`.


bjendres:
thanks @pfigel

PR #594, opened 2020-09-14
Improve installation guide

### LINKED ISSUE #593: Cannot install in CiviCRM 5.29.0
In the last three days, I've been evaluating CiviCRM for a local association. So far, it looks pretty good. As the association resides in Germany, SEPA integration would be great to have to help automate membership fees' debit transfers etc. Unfortunately, I cannot find out how to install this extension.

I've got two test systems now. The first instance runs Backdrop CMS. Installing CiviCRM was not an issue at all, it's running fine. The suggested installation via the extensions catalog doesn't work for my instance, the extension doesn't even show up. I tried to install the extension manually but to no avail, it doesn't show up in CiviCRM.

I guessed it might be a compatibility issue, so I set up a second site with Drupal 8. The same issues can be found here, though.

Is the extension even compatible with my CiviCRM version? As far as I understand CiviCRM's extension system, an extension that e.g., supports 5.5 should run on 5.29 as well. If it is compatible, what could be the reason it doesn't show up in the extensions catalog?

I'm really lost right now, running out of ideas how to get the extension to finally run...

TheAssassin:
The previous wording was misleading and caused confusion to new users (both of CiviCRM and the extension; see #593 for instance). This commit attempts to refine the wording to help new users install the extension.

Closes #593.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -32,11 +32,17 @@ Find more documentation on http://wiki.civicrm.org/confluence/display/CRM/CiviSE
 * membership payments
 * automatic submission to the banks
 
+
 ## Installation
-Use the following steps to install CiviSEPA.
- * **Administer** > **System Settings** > **Extensions**
- * **Add New**
- * Find, download, and install
+
+This extension needs to be installed manually into CiviCRM. It is not (yet) available from the built-in extensions catalog.
+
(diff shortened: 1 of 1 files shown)


bjendres:
@TheAssassin thanks, I'll take it :)

PR #595, opened 2020-09-14
GP-4897 Adjust logging schema after payment instrument upgrade

pfigel:
This adjusts the logging schema after adding payment instrument columns to the creditor table.


bjendres:
Thanks @pfigel, I can't believe I forgot...

PR #605, opened 2020-12-09
Fix composer and maildev issue in CI

### LINKED ISSUE #604: CI (continuous integration) broken
It seems that something's wrong with the CI, we get:
```
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - cweagans/composer-patches is locked to version 1.6.5 and an update of this package was not requested.
    - cweagans/composer-patches 1.6.5 requires composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.

You are using Composer 2, which some of your plugins seem to be incompatible with. Make sure you update your plugins or report a plugin-issue to ask them to support Composer 2.
```

Here's the full log: [ci.log](https://github.com/Project60/org.project60.sepa/files/5608940/ci.log)

Tests still pass on my local environment.

pfigel:
This fixes some CI issues by forcing composer to use v1 and adding the `maildev` image to work around a problem during Drupal setup.

The `maildev` issue is unrelated to composer, but we've been seeing that one from time to time for other extensions, and it hit me while testing this fix. It usually causes build failures with "Unable to send e-mail. Contact the site administrator if the problem persists." during the `civibuild` run.

Fixes #604

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,10 @@
 version: 2.1
 commands:
+  setup:
+    description: "Setup environment"
+    steps:
+      - run: |
+          composer self-update --1
   install_civicrm:
     description: "Install CiviCRM"
     parameters:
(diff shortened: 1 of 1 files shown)


bjendres:
let's see if that does the trick, thanks @pfigel

bjendres:
...and it does! Great work!! That would've taken me ages to find/fix!

Issue #609, opened 2021-03-23
Creation of mandate references relies on unreliable MySQL internals


rthouvenin:
When creating a new mandate, the following query is used to guess the next row id and append it to the mandate reference:
`SELECT auto_increment FROM information_schema.tables WHERE table_schema='$database' and table_name='civicrm_sdd_mandate'`.

This is unsafe in several ways:
 - MySQL uses locks for the duration of a **statement** (not transaction) to determine the next auto-increment value, therefore querying the next value from a separate statement does not guarantee it will be the one used by the insert.
 - MySQL has several ways to determine the next value, and in one of them ('interleave') the values are not guaranteed to be sequential. Since MySQL 8.0, this interleave mode is the default (it is more efficient and more reliable)
 - Since MySQL 8.0, by default MySQL caches the table stats in information schema for 24 hours, and the `tables.auto_increment` column is part of these stats. 

As a result, after our migration to MySQL 8 we are not able to create new SEPA mandates because they fail on the unique constraint for the mandate reference...

bjendres:
Yes, thanks @rthouvenin. This code has been in there from the very first prototype, and was always an eyesore :) Would you be able to provide a PR with an alternative?

rthouvenin:
@bjendres Yes will work out something, hopefully very soon as this is quite urgent for us now :)

bjendres:
> Yes will work out something hopefully very soon 

Great!

> ... as this is quite urgent for us now :)

Sorry about that :/

MasterEmit:
This is still open? Just after upgrading everything I got some days later the reply, that our office could sometimes not create sepa mandates and today I realize, that this problem still exists :-(

bjendres:
> This is still open? 

Yes, because of [this](https://github.com/Project60/org.project60.sepa/pull/610#issuecomment-986285789) issue.

> Just after upgrading everything I got some days later the reply, that our office could sometimes not create sepa mandates and today I realize, that this problem still exists :-(

Sorry about that. Would you be able to help out with the PR (#610)?

PR #610, opened 2021-03-23
Use more reliable method to generate unique mandate reference 

rthouvenin:
Fix for #609

The default reference pattern is going to be longer and use alphanumeric characters. While it would be possible to make it shorter and look like the previous pattern it would require making some guesses on how many mandates have been generated by existing users of the extensions so I didn't follow that path.

MySQL says that `uuid_short` guarantees uniqueness across replicas if it is called less than 16M times per second on average, so I think we should be fine :)

### DIFF
--- a/CRM/Sepa/BAO/SEPAMandate.php
+++ b/CRM/Sepa/BAO/SEPAMandate.php
@@ -71,8 +71,8 @@ class CRM_Sepa_BAO_SEPAMandate extends CRM_Sepa_DAO_SEPAMandate {
         // If no mandate reference was supplied by the caller nor the customisation hook, create a nice default one.
         $dao = new CRM_Core_DAO();
         $database = $dao->database();
-        $next_id = CRM_Core_DAO::singleValueQuery("SELECT auto_increment FROM information_schema.tables WHERE table_schema='$database' and table_name='civicrm_sdd_mandate';");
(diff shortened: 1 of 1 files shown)


MasterEmit:
This really should be included in the next update. Otherwise we always have to patch these two lines by hand to have the sepa mandate functionality working.

bjendres:
I get your point (and thanks for the PR by the way), but doesn't that mean that all people that haven't implemented their own reference generation will silently get different mandate references with the next update? That, in turn, can cause a couple of issues down the chain, e.g. the pattern to detect SEPA payments in a bank statement might have to change.

Could you maybe add this algorithm as a configurable option and leave the default as is? And maybe add a user notification with the upgrade script that it's more reliable to switch to the new generator but that the new scheme could potentially have an impact on other processes? What do you think?

jaapjansma:
I have just tested this patch and can confirm it works.

@bjendres regarding your concern, this for installation where people don't use their own mandate reference generation.

bjendres (approved):
Looks good to me, and @jaapjansma's happy with it, too.

bjendres:
Hi @jaapjansma, I've released this in [1.12-alpha1](https://github.com/Project60/org.project60.sepa/releases/tag/1.12-alpha1), hope that helps.

jaapjansma:
Thanks!

Issue #620, opened 2022-01-17
Add Recurring Contribution ID as Output


jojowork:
Using the Formprocessor, I need the recurring contribution ID of a created SEPA-Mandate.

bjendres:
@jojowork Do you need this to be part of ``1.6``? If so, please create a PR asap.

bjendres:
released with ``1.6.1``

PR #621, opened 2022-01-17
[#620] Add Recurring Contribution ID as Output as Output of Formprocessor Action

jojowork:


bjendres:
That might be confusing, because in the case of a OOFF mandate, the contribution ID (not the recurring contribution ID) is returned. I'd suggest to either
 * call it "entity ID" (or something similar)
 * have two return values ``contribution_id`` and ``contribution_rcur_id`` so it's clear.

bjendres:
Thanks @jojowork

bjendres:
released with ``1.6.1``

Issue #623, opened 2022-01-29
Incorrect tag in xml file (/pain_008_001_02_OTHERID)


Detsieber:
For some reason, the <CdtrSchmeId> section contains a superfluous <Nm>-Tag.
This causes the xml file to be rejected.

Pull request is under way.

bjendres:
@Detsieber can you rule out that the change will break submissions for other people/banks?

bjendres:
Actually, see #589 - so we might want to offer both versions...

Detsieber:
Sorry, I haven't seen before that #589 is absolutely identical with this issue #623. 
And I am still sure that the fix proposed by wintermoor and myself will help: He has documented, that the change is necessary for GLS bank (which uses the same software most Volksbanken in Germany are using. And in my case it concerns the Sparkasse.

As I wrote at #589: 
> However, I am pretty sure that no bank will accept the original file!
> Xavier was wrong here, and I also do not see how the cited documentation would support that tag as this place.
> Also, I tested this with the SEPA XML Checker from Star-Finanz-Software, and they complain about that tag too.
> 
> Therefore, please merge my commit.

And I do not believe that a SEPA format that will never work should be kept.

tttp:
> Sorry, I haven't seen before that #589 is absolutely identical with this issue #623.

Given how time works, I'd say that #623 is identical to #589 ;)

> Xavier was wrong here, and I also do not see how the cited documentation would support that tag as this place

`    <xs:element maxOccurs="1" minOccurs="0" name="Nm" type="Max140Text"/>
`

This lines say you can have a Nm tag.

I do not remember for which bank it was added, it might be one that required to have Nm, it was too long ago ;)

In any case, it should be an optional but accepted according to that xls, a bank is wrong to reject it, a bank is wrong to make it mandatory, we are stuck in the middle and we somehow have to deal with actors badly implementing the standards

Anyway, @rthouvenin, are we still using this template? could you check with Rola if the xml without the Sn would work?

otherwise: let's remove, let's keep, let's make two versions,  not my fight ;)

bjendres:
Closed in favour of #627.

Detsieber:
I reopened that one.
Again: The sfirm software, which is used a reference by the Sparkasse group, explicitely says that this Nm-tag is illegal!

[...]

PR #624, opened 2022-01-29
Update transaction-details.tpl

Detsieber:
see issue #623


bjendres:
rejected because of #627

Issue #639, opened 2022-07-21
PAIN format "pain.008.001.02 without BIC and CdtrSchmId:Nm" doesn't work for Switzerland


Detsieber:
SEPA is supporting direct debits from debitor accounts in Switzerland. However, in these cases, other than in any other country, the postal address of the debitor has to be added.

This was already implemented in "pain.008.003.02 (EBICS-3.0 RCUR/CH)".

Therefore, in "pain.008.001.02 without BIC and CdtrSchmId:Nm", directly after 

`          <Nm>{$contribution.display_name}</Nm>`

the following code has to be inserted 

```
{if $contribution.ctry eq 'CH'}
          <PstlAdr>
            <Ctry>{$contribution.ctry}</Ctry>
            <AdrLine>{$contribution.street_address}</AdrLine>
            <AdrLine>{$contribution.postal_code} {$contribution.city}</AdrLine>
          </PstlAdr>
{/if}

```
After our ongoing test is completed, I will add a pull request.

bjendres:
But I understand that you want a mixed mode format, which we have tried to avoid. Our approach was mostly to have different creditors for Swiss people, and/or generally people from abroad.

Detsieber:
In all use cases I can imagine, we won't have different contribution pages for people from Switzerland. Therefore, all SEPA mandates have to be processed by the same sepa creditor, using the same PAIN format.

I don't see any reason why we shouldn't do so?

bjendres:
> I don't see any reason why we shouldn't do so?

One of the reasons we wanted to separate, is that you then have to force *all* donors to give an address, even though it's just needed for the swiss donors.

bjendres:
...and I'm not sure that your mixed file would be valid XML, but I might be wrong....

Detsieber:
The address will be generated only, if $contribution.ctry eq 'CH'.

Anyhow, I just patched the affected production system, and we will see...

jensschuppe:
@Detsieber Could you please file your patch as a PR for a new separate format.

Detsieber:
@jensschuppe I still don't think it would be necessary to create a new format, for the following reasons:

1. The current format is disfunctional, since it is not possible to process sepa mandates from switzerland without adding the contributor's address (which is required for Switzerland). The proposed change will only add the contributor's address if the contributor is from Switzerland.
2. We have tested my PR in a pretty large production environment for2,5 months, everything works well.
[...]

PR #654, opened 2022-10-19
Fix for #639

Detsieber:
This is a fix for #639.

This is tested in production and works well.

### DIFF
--- a/templates/Sepa/Formats/pain_008_001_02_NOCSNM/transaction-details.tpl
+++ b/templates/Sepa/Formats/pain_008_001_02_NOCSNM/transaction-details.tpl
@@ -62,6 +62,13 @@
         </DbtrAgt>
         <Dbtr>
           <Nm>{$contribution.display_name}</Nm>
+          {if $contribution.ctry eq 'CH'}
+           <PstlAdr>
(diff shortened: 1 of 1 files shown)


Detsieber:
Would be nice to include into 1.7

bjendres:
Hi @Detsieber, thanks for the PR. But, as stated in https://github.com/Project60/org.project60.sepa/issues/639#issuecomment-1268431566, please adjust the PR so it creates a _new file format_ with the desired properties. This way we won't jeopardise the submissions of people using the file format you're changing...

bjendres:
remark: the commits https://github.com/Project60/org.project60.sepa/commit/21ddc472228a1a1645d093d365790b6aa34a1715 and https://github.com/Project60/org.project60.sepa/commit/9340a3931a7e906c35220ee4b49ce747c2f92510 mentioned above do *not* belong to this issue.

bjendres:
FYI, we solved this problem with _two_ creditors: one for switzerland with an format that has the address in the file, and another one for all other countries. I think the SEPA-Tools extension allows you to define the creditor based on a regex of the IBAN...

PR #664, opened 2023-04-13
Add CBIBdySDDReq_00_01_00 SEPA format

masetto:
The CBIBdySDDReq is a XML format used by Italian banks and requires a new setting field called "CUC-code".


bjendres (changes requested):
The PR looks pretty solid, thank you!

I'd only ask you to make it (even) more clear in the UI that this value is only used in this file format (so far), so we don't get a lot of questions about what to put there from 99% of the users that don't need it. Maybe a placeholder text saying "only for CBID format" or something would be enough. What do you think?

And, if you're feeling like going the extra mile today, maybe add a unit test for this file format.

masetto:
In the CUC field I removed the default value 'SEPATEST' and added a clarification in the label. Is it ok for you?
I do not know how to add a placeholder....

For unit tests please give me a starting point to take inspiration from?

bjendres:
> For unit tests please give me a starting point to take inspiration from?

You're right. I thought we already had some, but I think I was confusing that with another project. I created #665 to not forget about this. But I don't expect you to do that yourself, unless you're really motivated.

I think the next step would be for me to try this out locally and see if I can spot any more issues. 

If everything goes well, I'll merge and tag a 1.8-alpha version for you.

bjendres:
@masetto I have merged your PR and tagged it as [1.9-alpha1](https://github.com/Project60/org.project60.sepa/releases/tag/1.9-alpha1). 

Could you please confirm that this works? Specifically in the "newly installed" and the "upgrade from ``1.8.x``" scenarios?

masetto:
of course :smiley: we have already sent contributions to the bank.

PR #668, opened 2023-04-25
fix #667: set account holder to be an optional argument instead of re…

### LINKED ISSUE #667: Action create sepa mandate failed. Caused by account_holder is required
We updated CiviSepa to version 1.8 and got the following error message with formerly working formprocessor action: create recurring sepa mandate.

the error seems to be caused by this line:
https://github.com/Project60/org.project60.sepa/blob/1.8.0/Civi/Sepa/ActionProvider/Action/CreateRecurringMandate.php#L52 

The author seems to have unintendedly set the parameter to be required (and added a comment to later check what that flag means).

It also does not make a lot of sense to require an account holder field, because account holder is only an optional additional feature  for the case that the account holder differs from the name of contact.

I will create an PR that sets the flag to false (not required).

jojowork:
…quired

fixes issue #667


jojowork:
tested on one environment and working as expected.
@bjendres can you merge it?

PR #672, opened 2023-05-30
Fix NextCollectionDate updating non-SEPA ContributionRecurs

pfigel:
This fixes an issue where `CRM_Sepa_Logic_NextCollectionDate` may update `next_sched_contribution_date` for non-SEPA ContributionRecur entities when they are modified.


bjendres:
Thanks @pfigel

Issue #681, opened 2023-08-21
BIC is required in Actions to create Recurring or One Off but should respect the setting uses BIC for the account


ErikHommel:
At the moment the BIC is a required parameters in the Actions CreateOneOffMandate and CreateRecurringMandate. This does not respect the setting of the account, where uses BIC can be switched off. Should be changed to respect the setting.

bjendres:
Yes, it looks like this scenario was not considered in the implementation.

@ErikHommel Will you be working on this, or "just" reporting?

PR #690, opened 2023-11-16
Fix for issue #681 made bic optional in the form processor create mandate and create one off mandate actions

jaapjansma:
**Before**

BIC is **required** in the form processor create recurring mandate and create one off mandate actions

**After**

BIC is **optional** in the form processor create recurring mandate and create one off mandate actions


jaapjansma:
This PR now also checks whether BIC is required on the creditor.

PR #677, opened 2023-07-12
adds new style token processing

AlainBenbassat:
I noticed that the Most_Recent_SEPA_Mandate tokens are not replaced anymore in version 5.58.1, because the hooks hook_civicrm_tokens and hook_civicrm_tokenValues are deprecated.

This PR adds the new of processing tokens, while retaining the legacy way. 

See also: https://docs.civicrm.org/dev/en/latest/framework/token/#defining-tokens for more information.

My attempt in this PR is to:
- only touch the token code
- make new token processing and legacy processing as similar as possible
- convert array to [] in the files I changed
- move as much code out of sepa.php
  - 3 methods were added to the BAO SEPAMandate
  - a new helper class SepaTokens was added for the new way of handling tokens
  - a new helper class SepaTokensDeprecated handles the legacy token processing (it has the same methods as SepaTokens)

PR #693, opened 2023-12-18
#677 Adds new style token processing

kainuk:
The same PR as, but rebased to the latest code base and implemented the suggestion of @bjendres

Here is the original description of @AlainBenbassat:

I noticed that the Most_Recent_SEPA_Mandate tokens are not replaced anymore in version 5.58.1, because the hooks hook_civicrm_tokens and hook_civicrm_tokenValues are deprecated.

This PR adds the new of processing tokens, while retaining the legacy way.

See also: https://docs.civicrm.org/dev/en/latest/framework/token/#defining-tokens for more information.

My attempt in this PR is to:

only touch the token code
make new token processing and legacy processing as similar as possible
convert array to [] in the files I changed
move as much code out of sepa.php
* 3 methods were added to the BAO SEPAMandate
* a new helper class SepaTokens was added for the new way of handling tokens
* a new helper class SepaTokensDeprecated handles the legacy token processing (it has the same methods as SepaTokens)

### DIFF
--- a/CRM/Sepa/BAO/SEPAMandate.php
+++ b/CRM/Sepa/BAO/SEPAMandate.php
@@ -900,8 +900,10 @@ class CRM_Sepa_BAO_SEPAMandate extends CRM_Sepa_DAO_SEPAMandate {
       return FALSE;
     }
 
-    $query = "SELECT contact_id, MAX(id) AS mandate_id FROM civicrm_sdd_mandate WHERE contact_id = $cid GROUP BY contact_id;";
-    $result = CRM_Core_DAO::executeQuery($query);
+    $query = "SELECT contact_id, MAX(id) AS mandate_id FROM civicrm_sdd_mandate WHERE contact_id = %1 GROUP BY contact_id;";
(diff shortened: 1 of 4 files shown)


kainuk:
I have tested the PR on version 5.43. The old-skool triggers are needed for this version. I have also tested it on 5.65.1.

kainuk:
The test fail, but the error does not come from a line of code in the extension, so I do not know how to prevent it.

bjendres:
> The test fail, but the error does not come from a line of code in the extension, so I do not know how to prevent it.

Yes, we need to fix the CircleCI config, see #694

kainuk:
Hi @bjendres, I have another customer that needs this tokens, and would be helped with a merge.

bjendres:
I have merged it, please verify that it works with ``1.10-alpha3``: [org.project60.sepa-1.10-alpha3.tar.gz](https://github.com/Project60/org.project60.sepa/files/13955758/org.project60.sepa-1.10-alpha3.tar.gz)

Issue #697, opened 2024-01-21
Compatibility with CiviCRM 5.69.2 / Smarty V3


Detsieber:
Starting with Release 5.69.0, CiviCRM supports the long awaited Smarty3 framework. Currently, Smarty3 support is optional, and can be switched off if extensions are incompatible. However, extensions should be updated soon, as it is announced that Smarty2 support will end during the lifetime of CiviCRM 5.69ESR.

If Smarty3 is activated, CiviCRM produces php errors when trying to access contributions:

```
[error] 
$Fatal Error Details = array:3 [
  "message" => "Unable to load template 'file:CRM/Contact/Page/View/Summary.sepa.tpl'"
  "code" => null
  "exception" => SmartyException {#16911
    #message: "Unable to load template 'file:CRM/Contact/Page/View/Summary.sepa.tpl'"
    #code: 0
    #file: "/html/wordpress/wp-content/plugins/civicrm/civicrm/packages/smarty3/vendor/smarty/smarty/libs/sysplugins/smarty_internal_template.php"
    #line: 195
    trace: {
      /html/wordpress/wp-content/plugins/civicrm/civicrm/packages/smarty3/vendor/smarty/smarty/libs/sysplugins/smarty_internal_template.php:195 { …}
      /html/wordpress/wp-content/plugins/civicrm/civicrm/packages/smarty3/vendor/smarty/smarty/libs/sysplugins/smarty_internal_templatebase.php:232 { …}
      /html/wordpress/wp-content/plugins/civicrm/civicrm/packages/smarty3/vendor/smarty/smarty/libs/sysplugins/smarty_internal_templatebase.php:116 { …}
      /html/wordpress/wp-content/plugins/civicrm/civicrm/CRM/Core/Region.php:81 {
        CRM_Core_Region->{closure}
        › $smarty->assign('snippet', $snippet);
        › $html .= $smarty->fetch($snippet['template']);
        › $smarty->assign('snippet', $tmp);
        arguments: {
          $template: "CRM/Contact/Page/View/Summary.sepa.tpl"
        }
      }
      /html/wordpress/wp-content/plugins/civicrm/civicrm/CRM/Core/Region.php:157 { …}
      /html/wordpress/wp-content/plugins/civicrm/civicrm/CRM/Core/Smarty/plugins/block.crmRegion.php:26 { …}
[...]

PR #698, opened 2024-01-23
Fix for #697: Compatibility with CiviCRM 5.69

jaapjansma:
This fixes the issue with CiviCRM 5.69.

This PR is tested on CiviCRM 5.69 and 5.14 (so it is still backwards compatible)l

See #697


jaapjansma:
@bjendres are you able to review this PR and merge it if ok?

jensschuppe:
I think this should do a complete `civix upgrade`@`23.02.1` which leaves `_sepa_civix_civicrm_config()` with no template-specific code altogether …

jaapjansma:
Yes that might be good. However I have a little objection against that it might break the working of this extension on older versions of CiviCRM when we do that.

bjendres:
> @bjendres are you able to review this PR and merge it if ok?

I can't really assess the consequences of these changes, but it looks reasonable for me. If the two of you think it makes sense, I'll merge and release with ``1.10``.

bjendres:
Thanks @jaapjansma

Issue #655, opened 2022-10-26
Queue runner for 'group received' and 'delete group' actions


bjendres:
The submission group **update** and the group **closing** process already have a queue runner to avoid PHP timeouts when the groups are large. 

However, the same is needed for **deleting** groups and for marking them as **received**. Both cases also need a number of API calls that grows with the group size, and is therefore equally prone to PHP timeouts.

pminf:
We also have to deal with this problem because we regularly have to close groups of around 8000 entries. Unfortunately, the user does not receive any feedback if the request generates a timeout. You only notice it within the group and that the status of the individual contributions no longer changes.

bjendres:
As a workaround , you can close large groups via command line (i.e. no timeout) with
``
cv api3 SepaAlternativeBatching.received txgroup_id=XX
``

That *might* also work for already received, incomplete groups, but please test well, I'm not sure about this one.

pminf:
OK, good to know. But only admins have access to the cli, so it's not a practicable workaround for us.

bjendres:
> OK, good to know. But only admins have access to the cli, so it's not a practicable workaround for us.

I hope I can get to implementing this soon.

bjendres:
@pminf Could you test if this version (https://github.com/Project60/org.project60.sepa/commit/c73e7133b1ce56561d6ffd06bc6c7918cee58336) works for you? If you have large SEPA group support enabled, the "mark received" should now also be executed by a runner.

pminf:
@bjendres Works like charm: I've closed a group with over 7000 entries in one go without failure. Thanks!

PR #700, opened 2024-03-12
#655 Created "mark group received" runner

bjendres:
This additional runner will avoid timeout issues when marking large SEPA groups as "received".

*systopia-reference: 19531*

### DIFF
--- a/CRM/Sepa/Logic/Queue/Close.php
+++ b/CRM/Sepa/Logic/Queue/Close.php
@@ -1,7 +1,7 @@
 <?php
 /*-------------------------------------------------------+
 | Project 60 - SEPA direct debit                         |
-| Copyright (C) 2017-2018 SYSTOPIA                       |
+| Copyright (C) 2017-2024 SYSTOPIA                       |
 | Author: B. Endres (endres -at- systopia.de)            |
 | http://www.systopia.de/                                |
 +--------------------------------------------------------+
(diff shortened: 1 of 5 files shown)


bjendres:
@jensschuppe Could you please have a brief look to see you can spot any major blunders?

CRM/Sepa/Logic/Queue/Close.php
```diff
@@ -56,23 +63,30 @@ public static function launchCloseRunner($txgroup_ids, $target_group_status, $ta
...
       $contribution_count += SDD_CLOSE_RUNNER_BATCH_SIZE; // security margin
-      for ($offset=0; $offset <= $contribution_count; $offset+=SDD_CLOSE_RUNNER_BATCH_SIZE) {
+      for ($offset=0; $offset <= $contribution_count; $offset += SDD_CLOSE_RUNNER_BATCH_SIZE) {
         $queue->createItem(new CRM_Sepa_Logic_Queue_Close('update_contribution', $txgroup, $target_contribution_status, $offset));
       }
 
       // finally: render XML and mark the group
-      $queue->createItem(new CRM_Sepa_Logic_Queue_Close('create_xml', $txgroup, $target_group_status));
+      if ($is_received_runner) {
+        $queue->createItem(new CRM_Sepa_Logic_Queue_Close('create_xml', $txgroup, $target_group_status));
```
jensschuppe, line 72: According to the [docs](https://docs.civicrm.org/dev/en/latest/framework/queues/#2-create-itemstasks-on-the-queue), a queue should be filled with `CRM_Queue_Task` objects when it is supposed to be run using CiviCRM's queue runners. Not sure how much BC there is for the concept being used here, though …

jensschuppe, line 72: Adding items should only be done when the queue is empty or doesn't exist. This should be checked with `$queue->existsQueue()` before adding items. Otherwise, loading the `CRM_Sepa_Page_MarkGroupReceived` page again would load the existing queue and add items for all contributions again, and also starting the runner again (which might stall either runner due to the next item being already claimed by the other).

jensschuppe (commented):
I know this is old code, so some remarks might be negligible. But if there's time and budget, some refactoring would make sense.

bjendres:
> I know this is old code, so some remarks might be negligible. But if there's time and budget, some refactoring would make sense.

I agree, but the version as-is  has been tested by 3 clients now, and I'm going to merge it the way it is. I created #703 for your suggestions.

PR #701, opened 2024-03-13
civix upgrade to support Smarty3

pradpnayak:


bjendres:
Thanks, @pradpnayak. 

I'll assign this to ``1.11.x``, so we won't lose all clients still running CiviCRM < ``5.34``

jaapjansma:
I close this PR as this already fixed in this commit: https://github.com/Project60/org.project60.sepa/commit/397a069a079a17c42a8e32fd1882ff1a78e6d1ad

Issue #619, opened 2021-11-08
Deferring collection date after weekend not sufficient


Detsieber:
When the option "exclude_weekends" is selected, the collection date will be deferred to Monday, in case it would be on Saturday or Sunday.

However, this won't be sufficient in many cases:

Typically, the rules of several banks are as follows: 
- If the SEPA file is uploaded until 14:00h, the collection can happen the following day (means: CiviSEPA should be set up to defer the collection by 1 day)
- If the SEPA file is uploaded after 14:00h (which will happen typically if the process is done by volunteers after their regular working hours!), the collection has to be deferred by 2 days.

If in the latter case, the SEPA file gets uploaded on Friday afternoon (or Saturday), the collection date will be calculated by CiviSEPA to Monday. However, the bank will not accept this file, because they cannot collect earlier than Tuesday.

**Possible solution:**
There should be another setting besides the "exclude weekends": There one should be able to enter the number of extra_defer_days after a weekend (in this case: 1).

bjendres:
You can implement such an algorithm using the ``civicrm_defer_collection_date`` hook, see [HERE](https://github.com/Project60/org.project60.sepa/blob/1.6-alpha2/org.project60.sepacustom/sepacustom.php#L131-L146). 

That being said, it would be nice to have a reference implementation with configuration options for what you're describing, so anyone can use it. However, I would place this rather in a companion extension like [de.systopia.sepacustom](https://github.com/systopia/de.systopia.sepacustom) than in the CiviSEPA core.

pminf:
Hey @Detsieber, we have the same requirement and would like the collection date to be deferred to Tuesday, if the collection was submitted on a Friday. Have you used the mentioned hook to solve this? I would be grateful if you could share your solution here.

Detsieber:
Hi @pminf unfortunately, we haven't yet implemented this ... Making the calculation of the deferral "waterproof" was not so easy at first... and I did not give enough priority for that until now.

PR #704, opened 2024-03-26
Upgrade Civix-generated code to Civix version 23.02.1

jensschuppe:
This fixes e. g. the contact view failing with `Unable to load template 'file:CRM/Contact/Page/View/Summary.sepa.tpl'`.

I think this is related to *Smarty 3*, but haven't tested without it. This is on CiviCRM `5.71`. This PR raises CiviCRM minimum version from `5.43` to `5.45` due to use of the `entity-types` mixin.

If version `1.10` is supposed to stay in alpha for a while, we should consider merging this for a `1.9.x` version.


CRM/Sepa/Upgrader.php
```diff
@@ -21,7 +21,7 @@
 /**
  * Collection of upgrade steps.
  */
-class CRM_Sepa_Upgrader extends CRM_Sepa_Upgrader_Base {
```
jaapjansma, line 24: Not sure if this change is compatible with CiviCRM 5.45. I found this change breaking in older installations because it expects the base class to be present in CiviCRM core. Which is a wrong assumption if the extension needs to be compatible with older CiviCRM versions.

info.xml
```diff
@@ -24,14 +24,18 @@
...
   <civix>
     <namespace>CRM/Sepa</namespace>
-    <format>22.05.2</format>
+    <format>23.02.1</format>
   </civix>
   <mixins>
     <mixin>menu-xml@1.0.0</mixin>
     <mixin>mgd-php@1.0.0</mixin>
     <mixin>setting-php@1.0.0</mixin>
+    <mixin>smarty-v2@1.0.1</mixin>
```
jaapjansma, line 37: The update to 23.02.1 removes the mixin files from the extension. Which then would expect the functionality to be present in CiviCRM core. So again I am not sure whether that is the case with CiviCRM 5.45

jaapjansma (approved):
I approve with the changes. However I am not 100% confident whether all those changes are compatible with CiviCRM 5.45

CRM/Sepa/Upgrader.php
jensschuppe: The core class `CRM_Extension_Upgrader_Base` was added in CiviCRM `5.38`, so this should be fine. *Civix* is checking that and adjusting the `info.xml` accordingly.

info.xml
jensschuppe: This is also being checked by *Civix* and the `info.xml` adjusted accordingly. E.g. the `menu-xml` mixin, this got added to CiviCRM core with version `5.45` (see https://github.com/civicrm/civicrm-core/commit/7eab943a07f6d2fbc78e0f5ff062b52415f19a52).

CRM/Sepa/Upgrader.php
jaapjansma: Ok thanks! I did not know that

jaapjansma:
I think this merge ready

info.xml
jaapjansma: ok cool.

bjendres:
Should I tag ``1.10-alpha6``? I still want to implement #619 with ``1.10``, so no beta version just yet...

PR #708, opened 2024-05-01
civix-23.02.1 for smarty3

sluc23:
we still in branch 1.9.x for CiviCRM 5.69 and smarty3, we need to update civix.


jaapjansma:
I close this PR as this already fixed in this commit: https://github.com/Project60/org.project60.sepa/commit/397a069a079a17c42a8e32fd1882ff1a78e6d1ad

PR #710, opened 2024-05-07
Add html metadata to Mandate entity & regenerate DAOs

colemanw:
This adds some metadata so the Mandate entity can be used with FormBuilder. I recommend doing a similar update for the other entities. See discussion: https://chat.civicrm.org/civicrm/pl/jwtjgiyak7ddunzujnqfctbbzh

Note: the diff is noisy because of the `civix generate:entity-boilerplate` but the only functional change is the addition of the html => type metadata to the Mandate entity.


bjendres:
Great, thanks! Any implications for the minimum CiviCRM version?

bjendres:
@colemanw Sounds great, thanks! Do you think it's safe enough to still adding it to the ``1.10``, which has been in beta for some time?

colemanw:
> @bjendres Sounds great, thanks! Do you think it's safe enough to still adding it to the 1.10, which has been in beta for some time?

I believe this PR is completely safe, as it only adds a bit of metadata without changing any functionality.

bjendres:
Released with ``1.10-beta4``. Will create a ``1.10.0`` in about a week's time.

peth-systopia:
> Thanks, @colemanw
> 
> @bjendres I was preparing some reports with SearchKit and I had come across this limitation. With this change (and maybe some others) I think you could consider packing 3 SearchKit reports that could replace or complement the existing CiviReports. At the level of user experience, it would improve significantly.

@francescbassas Sounds good. Unfortunately, at the moment we cannot make that kind of improvements without funding. If you can provide some (which would be greatly appreciated), please open a new issue and get in touch with us.

bjendres:
> @francescbassas Sounds good. Unfortunately, at the moment we cannot make that kind of improvements without funding. If you can provide some (which would be greatly appreciated), please open a new issue and get in touch with us.

@colemanw's PR has already been merged, (almost) no additional work for us.

colemanw:
> @francescbassas I was preparing some reports with SearchKit and I had come across this limitation. With this change (and maybe some others) I think you could consider packing 3 SearchKit reports that could replace or complement the existing CiviReports. At the level of user experience, it would improve significantly.

Now that the new version is released with this patch, try upgrading your extension and then creating the SearchKit report. If you get something useful, you could use the `civix export` command to add it to this extension & submit a PR :)
If you're not sure how to do that I can walk you through the steps, it's pretty straightforward.

peth-systopia:
@bjendres I was referring to the proposal to ship some packaged SearchKits with the extension.

bjendres:
> @bjendres I was referring to the proposal to ship some packaged SearchKits with the extension.

Ah, got it. Sure, sounds like a good idea. Could you create a separate ticket for that?

francescbassas:
I just updated to 1.10.0 and tried FormBuilder mandate fields integration and it works perfect! Thanks again.

Issue #715, opened 2024-05-17
Mandate date without Iban causes email crash utilizing Most_Recent_SEPA_Mandate token


kainuk:
First, this is the error: 

    TypeError: Cannot access offset of type string on string in Civi\Token\TokenRow->tokens() (line 171 of /var/www/html/vendor/civicrm/civicrm-core/Civi/Token/TokenRow.php).

To reproduce 

1) Create a contact
2) Create an OOFF mandate for this contact with a valid IBAN.
3) Remove the IBAN from the database using SQL (You can search with the SEPA Mandate code).

Try to send this contact and email.

The above error is thrown.

A good question is, does an OOFF mandate make sense without IBAN.  I do not think so. And I would not know how to create such a mandate from the web interface. But the database permits the column to be NULL. And I found a small number records in the database of a client.

bjendres:
Thanks, PR #716 merged

PR #716, opened 2024-05-17
Fix for #715 Mandate date without Iban causes email crash utilizing Most_Recent_SEPA_Mandate token

kainuk:


bjendres:
Thanks, @kainuk, seems harmless enough to include in the upcoming ``1.10``

PR #720, opened 2024-05-27
Setting SepaMandate API4 permissions

francescbassas:


bjendres:
@dontub could you have a quick look please?

dontub:
Without knowing this extensions that much, it makes sense to me, unless users should not be able use this API entity to make changes. Maybe the create and delete permissions defined in https://github.com/Project60/org.project60.sepa/blob/6e6c08464aa293f289a586aef1f43715851670a8/sepa.php#L362 might also be considered here... Additionally the `meta` permission might be limited to users that have one of the permissions defined in the linked method. Though without knowing the extension in detail I cannot ultimately judge that.

Issue #721, opened 2024-05-31
OOFF Contribution Status change batch fails 


kainuk:
After sending the xml file with OOFF transactions, successfully to the bank, the batch that updates the status from Pending to Complete fails. The batch stops in the screen below.

![2024-05-31_08-23](https://github.com/Project60/org.project60.sepa/assets/14834891/a74a9308-9689-4b70-8b03-6e66d873c6a6)

In the background an error is thrown.

````
TypeError: implode(): Argument #1 ($pieces) must be of type array, string given in implode() (line 187 of /var/www/html/web/sites/default/files/civicrm/ext/org.project60.sepa/CRM/Sepa/Logic/Queue/Close.php).
````

I have this error on the latest version of the sepa code with a PHP 8.1 install and CiviCRM 5.67 on Drupal 10.

pminf:
I can confirm this issue with latest CiviCRM 5.74.1, also using PHP 8.1.

In our case `\CRM_Sepa_Logic_Queue_Close::$origin_status_ids` is null, which leads to an error with the implode function.

![ooff-debug](https://github.com/Project60/org.project60.sepa/assets/611165/56c9b126-be3d-477a-b4e1-83f6c321b4e4)

kainuk:
Hi @bjendres, I have made a PR, see https://github.com/Project60/org.project60.sepa/pull/722 , but I would like feedback on it.

Detsieber:
I had this same issue on a large install (CiviCRM 5.74.4, CiviSEPA 1.10.0(!)), and this fix (#722) is working for us.
 Thank you @kainuk so much!

Please merge.

peth-systopia:
As @kainuk also indicates, this ist not (yet) a quick merge. So still 'needs funding'.

Detsieber:
I don't know under which conditions this error occurs - but it breaks CiviSEPA completely!

IMHO, releasing a fix is important.

peth-systopia:
> IMHO, releasing a fix is important.

Can we count on you for some funding then?

pminf:
We've tested https://github.com/Project60/org.project60.sepa/releases/tag/1.11-alpha1 including this fix successfully.

GFF-DS:
Seems as if with [Release 1.11.0](https://github.com/Project60/org.project60.sepa/releases/tag/1.11.0) this issue is fixed and can be closed?

PR #722, opened 2024-05-31
Fix or issue 721

kainuk:
Fix for #721 , but it is done by restoring old code. So maybe I am missing something that could by fixed by changing something in the configuration.

What is a bit of a mystery for my is that the batch fails because the variable ` $this->origin_status_ids` is `NULL` but I have no idea where it can be set.


Detsieber:
I had this same issue on a large install (CiviCRM 5.74.4, CiviSEPA 1.10.0(!)), and this fix is working for us.
 Thank you @kainuk so much!

Please merge.

peth-systopia:
> Fix for https://github.com/Project60/org.project60.sepa/issues/721 , but it is done by restoring old code.

Not sure this qualifies for merging without further investigation, so I added the 'needs funding' tag.

MarcMichalsky:
> Not sure this qualifies for merging without further investigation

I have looked at the [latest changes](https://github.com/Project60/org.project60.sepa/commit/c73e7133b1ce56561d6ffd06bc6c7918cee58336) before the problem appeared. A method `setOriginStatusIds()` was added that is supposed to set `$this->origin_status_ids`, but I can't find any place in the repository where this method is used. Looks like a trivial mistake to me. The change by @kainuk just restores the old functionality.

Perhaps the original intention of the change should still be corrected and finalised, but the fix certainly helps to make the extension usable again in the meantime. 

Thanks for fixing and merging!

Issue #724, opened 2024-07-15
Add source to action to create One Off and Recurring Mandates


ErikHommel:

ErikHommel:
Also add creation date

PR #725, opened 2024-07-15
fix issue #724 - add source to create mandate actions

ErikHommel:


ErikHommel:
Also add creation date as a parameter (PR funded by Vluchtelingenwerk Vlaanderen)

PR #726, opened 2024-08-01
Error: Failed opening required 'CRM/Sepa/CustomData.php'

bjendres:
Remove path from CustomData ``require``


jensschuppe:
Actually, the class loader takes care of including PHP files, the `require_once` statement should thus be obsolete in the first place.

bjendres:
> Actually, the class loader takes care of including PHP files, the require_once statement should thus be obsolete in the first place.

Yes, but without it I'm getting the error above on my local *and* our HostShare environments.

bjendres:
Seems to work for me

PR #730, opened 2024-08-14
Replace deprecated `${}` syntax for referencing variables in strings

jensschuppe:


bjendres:
@jensschuppe would this cause problems in current systems, i.e. is this severe enough to shoehorn this into the ``1.11`` release?

jensschuppe:
No, the "new" syntax has been supported by PHP from the beginning, see [here](https://php.watch/versions/8.2/$%7Bvar%7D-string-interpolation-deprecated).

> The replacement examples mentioned above are compatible with all PHP versions.

bjendres:
@dontub Can you see *any* issues with this change? It needs to be rock solid if we want merge it this late in the beta-phase...

PR #732, opened 2024-08-19
Fix `ts()` calls with variable strings

jensschuppe:
Those strings haven't been part of the translation template, so translations won't be affected.


bjendres (approved):
looks ok.

PR #733, opened 2024-08-19
Add SYSTOPIA extension template with QA tools and GitHub actions

jensschuppe:
phpcs and phpstan GitHub actions are set to only run on `workflow_dispatch` due to too many errors.


dontub:
I'm afraid if the actions only run `workflow_dispatch` the rules won't be regarded for new code. Have you considered to limit phpcs and phpstan to the `Civi` directory?

jensschuppe:
Actually, the point was to be able to already include the desired rules but not run them on GitHub for now, until the code can be cleaned up. Restricting to only small parts of code will disable all checks locally as well (which I don't find useful tbh).

`workflow_dispatch` allows you to manually run things on GitHub when desired. But you are sure, this requires every contributor or reviewer to run the checks locally for QA. I'm not sure what's the better approach …

dontub:
I agree, restricting the checks to only a smaller part is definitely not ideal. Though I'm afraid if one runs `composer phpcs` or `composer phpstan` and gets a bunch of errors it might not be really helpful and will be ignored.

It is always possible to add paths in a local configuration file, if required...

jensschuppe:
Sure, feel free to try that in the `phpstan.neon.template` file. I'm not married to either approach :smirk:

dontub:
Closed in favor of #794.

PR #736, opened 2024-12-29
Towards supporting Smarty5 (replace addDate by datepicker)

mlutfy:
Fixes a few jcalendar/smarty5 issues on the "Create new Mandate" form, but it still has a bug, which unfortunately I could not figure out.

To reproduce: enable smarty5 (by default in CiviCRM 5.80+), enable SEPA ext, and then create a new mandate. The popup will fail with a network error, and the error will be `Syntax error in template "file:CRM/common/jcalendar.tpl" `.

This PR includes:

- Replaces `$form->addDate` by `$form->add('datepicker', ...`
- Removes `jcalendar` from the tpl files (it does not work on smarty5)
- Uses `CRM.utils.formatDate()` for formatting dates
- Fixes the JavaScript closure so that we can use `$` and `ts`  without the `domain` (but didn't change everything)

This PR has a bug:

-  `sdd_recalculate_fields` recalculates all fields, and also updates them, which causes recursive updates if we want to trigger "change" so that things like selecting the earliest rcur start date works. I don't know enough about the logic of the code to be sure that I will not be introducing bugs.

### DIFF
--- a/js/CreateMandate.js
+++ b/js/CreateMandate.js
@@ -12,7 +12,8 @@
 | written permission from the original author(s).        |
 +--------------------------------------------------------*/
 
-cj(document).ready(function() {
+(function($, _, ts) {
+  $(document).ready(function($) {
     /**
      * Identify and get the jQuery field for the given value
      */
@@ -30,7 +31,9 @@ cj(document).ready(function() {
 
         // flash the field a little bit to indicate change
         sdd_getF(fieldname).parent().fadeOut(50).fadeIn(50);
-        sdd_recalculate_fields();
+        if (recalculate) {
+            sdd_recalculate_fields();
+        }
     }
 
     /**
@@ -38,14 +41,8 @@ cj(document).ready(function() {
      *  using the sdd_converter element
      */
     function sdd_formatDate(date) {
-        cj("#sdd-create-mandate")
-            .find("[name^=sdd_converter].hasDatepicker")
-            .datepicker('setDate', date);
-
-        return cj("#sdd-create-mandate")
-            .find("[name^=sdd_converter_display]").val();
+        return CRM.utils.formatDate(date);
     }
-
 
     // logic to hide OOFF/RCUR fields
     function sdd_change_type() {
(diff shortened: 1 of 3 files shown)


js/CreateMandate.js
```diff
@@ -24,29 +25,25 @@ cj(document).ready(function() {
      * Utility function to set the date on the %^$W#$&$&% datepicker elements
      * PRs welcome :)
      **/
-    function sdd_setDate(fieldname, date) {
+    function sdd_setDate(fieldname, date, recalculate = true) {
```
mlutfy, line 28: This `recalculate` param can be ignored, does not work. It was to avoid recursive issues, when `sdd_setDate` is called from `sdd_recalculate_fields`. It is not enough, because there is also an event listener for `change` on most of the date fields.

bjendres: Thanks a lot @mlutfy, we'll try to pick this up after the holidays.

Happy holidays and new year!

bjendres (approved):
Looks good to me, @mlutfy, thanks.

@jensschuppe could you have a look and test it?

js/CreateMandate.js
mlutfy: I removed the `recalculate` param, which didn't work, to make this PR a bit more clean.

mlutfy:
I cleaned up the PR a bit, removing the `recalculate` param (which didn't work). However, this bug remains:

> sdd_recalculate_fields recalculates all fields, and also updates them, which causes recursive updates if we want to trigger "change" so that things like selecting the earliest rcur start date works. I don't know enough about the logic of the code to be sure that I will not be introducing bugs.

(i.e. you probably do not want to merge this PR as-is)

mlutfy:
@dontub Sounds like a solution. Were you able to test if it works for you?

With this PR, I mostly wanted to help move things forward, but I would need someone else to take the lead. I don't actively use this extension.

dontub (approved):

dontub:
I added the suggested code to prevent endless recursion and made minor code style related changes.

mlutfy:
@dontub Thank you!

Issue #739, opened 2025-02-26
unknown modifier 'sprintf'


bwurst:
I have installed a new CiviCRM with CiviSEPA. After applying the patch from pull request #736 I could set up a sepa mandate.

When I click on "Close and Send" in CiviSEPA dashboard, another smarty error occures: Syntax error in template "file:CRM/Sepa/Page/CloseGroup.tpl" on line 22 "<p>{$group_text|sprintf:$txgroup.reference}</p>" unknown modifier 'sprintf'

I modified the specified line to simply replace "sprintf" with "string_format" to fix this issue.
(But there seems to be a bunch of smarty5 errors left...)

Fabian-SYSTOPIA:
Could you sepcify the CiviSepa and CiviCRM versions you are using?

bwurst:
Installed a few days ago, CiviCRM 5.82.0 and CiviSEPA 1.11.0 

(When I look at the extension details, CiviSEPA is not marked as compatible with 5.82 but there was no complaint when I set thing up. Does CiviCRM no check version compatibility?)

Fabian-SYSTOPIA:
Could you try if a more recent pre-release fixes your problem such as [1.12beta1](https://github.com/Project60/org.project60.sepa/releases/tag/1.12-beta1)? 

If not, maybe @bjendres could have a look?

bjendres:
@bwurst Thanks for reporting!

Could you check if changing
```
<p>{$group_text|sprintf:$txgroup.reference}</p>
```
to
```
<p>{$group_text|string_format:"%s"}</p>```
```
in [this line](https://github.com/Project60/org.project60.sepa/blob/6647aab403d545160d12f1eb0ce2f403ed0f484a/templates/CRM/Sepa/Page/CloseGroup.tpl#L22) fixes it for you?

bjendres:
(updated last comment)

bwurst:
Yes, changing the given line does fix this, as I stated directly in the bug report. :)

Upgrading to beta does not help.

bjendres:
> Yes, changing the given line does fix this, as I stated directly in the bug report. :)

Just wanted to make sure I'm not forgetting anything.

Should be released with next ``1.12`` (#740) and next ``1.13`` (#741)

PR #740, opened 2025-02-27
#739 replace sprintf smarty modifier - 1.12.x

bjendres:
replaces discontinued ``sprintf`` modifier

### DIFF
--- a/templates/CRM/Sepa/Page/CloseGroup.tpl
+++ b/templates/CRM/Sepa/Page/CloseGroup.tpl
@@ -19,7 +19,7 @@
 			{capture assign=group_text}
 			<font size="+1">{ts domain="org.project60.sepa"}Group '%s' <b>is now closed</b> and cannot be changed any more.{/ts}</font>
 			{/capture}
-			<p>{$group_text|string_format:$txgroup.reference}</p>
+			<p>{$txgroup.reference|string_format:$group_text}</p>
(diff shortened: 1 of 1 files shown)


templates/CRM/Sepa/Page/CloseGroup.tpl
```diff
@@ -19,7 +19,7 @@
 			{capture assign=group_text}
 			<font size="+1">{ts domain="org.project60.sepa"}Group '%s' <b>is now closed</b> and cannot be changed any more.{/ts}</font>
 			{/capture}
-			<p>{$group_text|sprintf:$txgroup.reference}</p>
+			<p>{$group_text|string_format:$txgroup.reference}</p>
```
dontub, line 22: Should be `$txgroup.reference|string_format:$group_text` because `$group_text` contains the format.

See https://www.smarty.net/docs/en/language.modifier.string.format.tpl

dontub (changes requested):

templates/CRM/Sepa/Page/CloseGroup.tpl
dontub, line 22:
```suggestion
			<p>{$txgroup.reference|string_format:$group_text}</p>
```

bjendres (commented):
looks good to me

PR #744, opened 2025-03-20
Avoid parallel async batch run

### LINKED ISSUE #738: Parallel batching exclusion fails
Under certain circumstances, the same SEPA collection group apparently can be generated twice or more. This was observed when the same batching process was started at roughly the same time from different user accounts.

* CiviSepa-version: observed on ``1.13-beta1`` but probably present before
* systopia-ref: 28126

A rough to implementation for a working mutex lock should be:

1. function ``is_batching_busy`` with a reasonable timeout
2. function ``process_batching_item`` processing the the packets, and extending the lock's timeout

dontub:
Fixes #738.

In case of very long async batch runs it was possible to have parallel batch runs. All locking is now done with `\Civi\Sepa\Lock\SepaBatchLock` which combines CiviCRM's locking mechanism (DB lock) with a modified version of the already available lock mechanism for async runs. See the implementation of `\Civi\Sepa\Lock\SepaBatchLock` and `\Civi\Sepa\Lock\SepaAsyncBatchLock` for details.

The locking mechanism is also applied when closing groups.

systopia-reference: 28126


CRM/Sepa/Logic/Queue/Update.php
```diff
@@ -116,32 +121,36 @@ protected function __construct($cmd, $mode, $creditor_id = NULL, $offset = NULL,
...
 
       case 'UPDATE':
-        if ($this->mode == 'OOFF') {
-          CRM_Sepa_Logic_Batching::updateOOFF($this->creditor_id, 'now', $this->offset, $this->limit);
+        if ($this->mode === 'OOFF') {
+          CRM_Sepa_Logic_Batching::updateOOFF($this->creditorId, 'now', $this->offset, $this->limit);
         } else {
-          CRM_Sepa_Logic_Batching::updateRCUR($this->creditor_id, $this->mode, 'now', $this->offset, $this->limit);
+          CRM_Sepa_Logic_Batching::updateRCUR($this->creditorId, $this->mode, 'now', $this->offset, $this->limit);
         }
```
jensschuppe, line 144: Let's apply code formatting while we're at it (`else` on new line, comparison value on left side).

CRM/Sepa/Page/MarkGroupReceived.php
```diff
@@ -52,25 +53,35 @@ function run() {
     }
 
     // run the 'mark received' process
+    $this->closeGroup($group_id);
+
+    // go back to the dashboard
+    CRM_Utils_System::redirect(CRM_Utils_System::url('civicrm/sepa/dashboard', 'status=closed'));
```
jensschuppe, line 59: Why is closing the group necessary here, only closed groups can be marked received …?

Civi/Sepa/Lock/SepaBatchLock.php
```diff
@@ -0,0 +1,93 @@
...
+    // There is a small chance that the CiviCRM lock is acquired during a page
+    // switch when doing an async run. Thus, we release the lock wait a short
+    // time to give a possible async run the possibility to acquire the lock
+    // again. If the lock isn't acquired then, there's no async update running,
+    // but the async lock wasn't released for some reason.
+    $this->civiLock->release();
+    sleep(3);
+    if (!$this->civiLock->acquire(0)) {
+      return FALSE;
+    }
```
jensschuppe, line 74: This looks a bit arbitrary … can we have a look at that together?

l10n/org.project60.sepa.pot
jensschuppe, line 1: Are the translation changes related to this PR?

l10n/de_DE/LC_MESSAGES/sepa.po
jensschuppe, line 1: Are the translation changes related to this PR?

dontub: I applied the `pot` file and also fixed some strings where the fuzzy translation wasn't correct. (There are still many strings that are part of the `pot` file, but which I've dropped from the `po` file to avoid a too large diff.)

CRM/Sepa/BAO/SEPAMandate.php
```diff
@@ -686,27 +672,18 @@ static function modifyMandate($mandate_id, $changes) {
...
-
-
     // finally: generate acitivity
-    if (!empty($changes_details)) try {
+    if (!empty($changes_details)) {
       if (count($changes_subjects) == 1) {
         self::generateModificationActivity($mandate, $changes_subjects[0], $changes_details);
       } else {
-        $subject = ts("Multiple changes", array('domain' => 'org.project60.sepa'));
+        $subject = ts('Multiple changes', ['domain' => 'org.project60.sepa']);
```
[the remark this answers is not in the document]
dontub: Changed.

CRM/Sepa/Logic/Queue/Update.php
dontub: Changed.

CRM/Sepa/Page/MarkGroupReceived.php
dontub: I've renamed the method to `markGroupAsReceived()`. (The code is the same as before, but with locking.)

Civi/Sepa/Lock/SepaBatchLock.php
dontub: Sure.

PR #745, opened 2025-03-21
Use `string_format` instead of `sprintf` in `CloseGroup.tpl`

dontub:
This replaces the Smarty modifier `sprintf` with [`string_format`](https://www.smarty.net/docs/en/language.modifier.string.format.tpl).

The usage of `sprintf` resulted in this error:

`Syntax error in template "file:CRM/Sepa/Page/CloseGroup.tpl" on line 22 "<p>{$group_text|sprintf:$txgroup.reference}</p>" unknown modifier 'sprintf'`


jensschuppe:
Duplicate of #741 and #740 - can you leave a review on them instead, @dontub?

PR #747, opened 2025-04-01
Improve locking

### LINKED ISSUE #738: Parallel batching exclusion fails

dontub:
The first approach in #744 to fix #738 didn't work as expected. In between the HTTP requests for every queue item it was still possible to acquire the CiviCRM lock when a queue was running. The `\Civi\Sepa\Lock\SepaAsyncBatchLock` now uses a file instead of the DB. This allows to check if the acquire time is changed. This wouldn't be possible with the previous usage of the DB during an open transaction.

Releasing the CiviCRM lock and checking if the async lock is acquired is now also working better.

systopia-reference: 28126

### DIFF
--- a/Civi/Sepa/Lock/SepaAsyncBatchLock.php
+++ b/Civi/Sepa/Lock/SepaAsyncBatchLock.php
@@ -118,7 +118,7 @@ final class SepaAsyncBatchLock {
         throw new \RuntimeException("Couldn't remove {$this->lockFilePath}");
       }
     }
-    else if (FALSE === file_put_contents($this->lockFilePath, $id)) {
+    elseif (FALSE === file_put_contents($this->lockFilePath, $id)) {
       throw new \RuntimeException("Couldn't write {$this->lockFilePath}");
     }
   }
--- a/Civi/Sepa/Lock/SepaAsyncBatchLock.php
+++ b/Civi/Sepa/Lock/SepaAsyncBatchLock.php
@@ -31,10 +31,10 @@ final class SepaAsyncBatchLock {
 
   private ?string $acquiredId = NULL;
 
-  private string $lockFile;
+  private string $lockFilePath;
 
-  public function __construct(string $lockFile) {
-    $this->lockFile = $lockFile;
+  public function __construct(string $lockFilePath) {
+    $this->lockFilePath = $lockFilePath;
   }
 
   public function acquire(string $id): bool {
@@ -61,14 +61,14 @@ final class SepaAsyncBatchLock {
    *   is free.
    */
   public function getAcquireTime(): ?int {
-    if (!file_exists($this->lockFile)) {
+    if (!file_exists($this->lockFilePath)) {
       return NULL;
     }
 
-    clearstatcache(TRUE, $this->lockFile);
-    $time = filemtime($this->lockFile);
+    clearstatcache(TRUE, $this->lockFilePath);
+    $time = filemtime($this->lockFilePath);
     if (FALSE === $time) {
-      throw new \RuntimeException("Couldn't get modification time of {$this->lockFile}");
+      throw new \RuntimeException("Couldn't get modification time of {$this->lockFilePath}");
     }
 
     return $time;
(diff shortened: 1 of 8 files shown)


Civi/Sepa/Lock/SepaAsyncBatchLock.php
```diff
@@ -83,30 +92,35 @@ public function release(string $id): bool {
...
-  private function setAsyncLocks(array $locks): void {
-    \CRM_Sepa_Logic_Settings::setGenericSetting($locks, 'sdd_async_batching_lock');
+  private function setAsyncLock(?string $id): void {
+    $this->acquiredId = $id;
+    if (NULL === $this->acquiredId) {
+      if (!unlink($this->lockFile)) {
+        throw new \RuntimeException("Couldn't remove {$this->lockFile}");
+      }
+    }
+    else if (FALSE === file_put_contents($this->lockFile, $id)) {
```
jensschuppe, line 121: `s/else if/elseif`

Civi/Sepa/Lock/SepaAsyncBatchLock.php
```diff
@@ -29,29 +29,30 @@
  */
 final class SepaAsyncBatchLock {
 
-  private string $name;
+  private ?string $acquiredId = NULL;
 
-  public function __construct(string $name) {
-    $this->name = $name;
+  private string $lockFile;
```
jensschuppe, line 34: `$lockFilePath` because it's not a file handle?

Civi/Sepa/Lock/SepaBatchLock.php
```diff
@@ -27,18 +27,28 @@
...
   public function __construct(LockInterface $civiLock, SepaAsyncBatchLock $asyncLock, int $defaultTimeout) {
+    self::$instanceCount++;
     $this->civiLock = $civiLock;
     $this->asyncLock = $asyncLock;
     $this->defaultTimeout = $defaultTimeout;
   }
 
+  public function __destruct() {
+    self::$instanceCount--;
+    if (self::$instanceCount === 0) {
```
jensschuppe, line 47: Check for `<= 0` in case the static counter has been decreased somewhere else in the class?

Civi/Sepa/Lock/SepaBatchLock.php
```diff
@@ -58,17 +68,31 @@ public function acquire(?int $timeout = NULL, ?string $asyncLockId = NULL): bool
...
+    /*
+     * It is possible that the CiviCRM lock is acquired when during a queue
+     * execution an HTTP request was finished and the next one is not running,
+     * yet. Thus, we release the lock and wait some time to give a queue item
+     * the possibility to acquire the lock again. If the async lock time isn't
+     * changed, but the async lock wasn't released for some reason.
+     */
     $this->civiLock->release();
-    sleep(3);
+    for ($i = 0; $i < 5; ++$i) {
```
jensschuppe, line 84: Had me look up again whether pre/post-incrementing in `for` loops has any effect … :nerd_face:

jensschuppe (changes requested):

jensschuppe (commented):
I think it makes sense documenting how the locking algorithm works in the extension docs.

Civi/Sepa/Lock/SepaBatchLock.php
dontub: There's no other location where this counter is decreased and there cannot be any other.

Civi/Sepa/Lock/SepaBatchLock.php
dontub: As it doesn't I can change it to `$i++` if that increases the readability...

Civi/Sepa/Lock/SepaAsyncBatchLock.php
dontub: Changed.

Civi/Sepa/Lock/SepaAsyncBatchLock.php
dontub: Changed.

Civi/Sepa/Lock/SepaBatchLock.php
jensschuppe: Well, theoretically … but okay, let's not overthink

jensschuppe (approved):

PR #748, opened 2025-04-04
Use only one loop for preparing manadates for RCUR batching

jensschuppe:
This might fix side effects of the previously re-used reference variable while looping through mandates (introduced with #731).

### DIFF
--- a/CRM/Sepa/Logic/Batching.php
+++ b/CRM/Sepa/Logic/Batching.php
@@ -113,7 +113,9 @@ class CRM_Sepa_Logic_Batching {
       ->execute()
       ->getArrayCopy();
 
-    foreach ($relevant_mandates as &$mandate) {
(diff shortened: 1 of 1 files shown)


dontub (approved):
It makes sense to combine the two loops.

> This might fix side effects [...]

I cannot see that this change could have any effect on the result.

dontub:
> I cannot see that this change could have any effect on the result.

I see, the last mandate is overridden here https://github.com/Project60/org.project60.sepa/blob/b77f32a073a715cfec1b7acfe1399152fc149342/CRM/Sepa/Logic/Batching.php#L143
because it is a reference: https://github.com/Project60/org.project60.sepa/blob/b77f32a073a715cfec1b7acfe1399152fc149342/CRM/Sepa/Logic/Batching.php#L116

PR #750, opened 2025-04-29
Fix date comparison in query for finding obsolete contributions when terminating mandates

jensschuppe:
Terminating mandates currently fails when constructing queries with `WHERE receive_date = 'today'` when trying to identify obsolete contributions. This PR formats the date for MySQL, but the query should probably be replaced with an APIv4 call instead.

Since this never seems to have been found it could probably use a test for all possible values for the date value passed to `\CRM_Sepa_BAO_SEPAMandate::terminateMandate()` in `$new_end_date_str`. The method's signature is not clear about what it expects.

This should be backported to `1.12`.


dontub (approved):

bjendres (approved):
looks good to me

PR #759, opened 2025-06-16
Use hook instead of overwriting `add()` in `CRM_Sepa_DAO_SEPAMandate`

dontub:
This allows to store custom values for `SepaMandate` (if corresponding option value in group `cg_extend_objects` exists).

Note: I'm wondering if `status` needs to be loaded for existing mandates if it is not part of `$params` [at line 110](https://github.com/Project60/org.project60.sepa/commit/477f1ba73735c33f21d541bed20ddf802edbb534#diff-cdd059fb691dcddc967eee350a2e629f11063e57c1e4bd793877eff47d10b8e4R110) for this code:
```php
if (self::is_active($params['status'] ?? NULL)) {
  $params['validation_date'] ??= date('YmdHis');
}
```

systopia-reference: 28482

### DIFF
--- a/CRM/Sepa/BAO/SEPAMandate.php
+++ b/CRM/Sepa/BAO/SEPAMandate.php
@@ -21,65 +21,65 @@
  *
  */
 
+use Civi\Api4\SepaCreditor;
+use Civi\Api4\SepaMandate;
+use Civi\Core\Event\PreEvent;
+use Civi\Core\HookInterface;
 use Civi\Sepa\Lock\SepaBatchLockManager;
 use CRM_Sepa_ExtensionUtil as E;
 
 /**
  * Class contains functions for Sepa mandates
  */
-class CRM_Sepa_BAO_SEPAMandate extends CRM_Sepa_DAO_SEPAMandate {
+class CRM_Sepa_BAO_SEPAMandate extends CRM_Sepa_DAO_SEPAMandate implements HookInterface {
 
   /**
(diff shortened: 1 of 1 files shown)


jensschuppe (changes requested):
The `pre` hook is called when deleting as well, we should explicitly check for relevant actions (create/update), as validating makes no sense when deleting.

I agree to loading the status if necessary, although the current implementation doesn't do that either.

CRM/Sepa/BAO/SEPAMandate.php
```diff
@@ -37,6 +37,10 @@ class CRM_Sepa_BAO_SEPAMandate extends CRM_Sepa_DAO_SEPAMandate implements HookI
    * @throws \CRM_Core_Exception
    */
   public static function self_hook_civicrm_pre(PreEvent $event): void {
+    if ('delete' === $event->action) {
```
jensschuppe, line 40: The old method got called for `create`, `save` and `update` actions (see [here](https://github.com/civicrm/civicrm-core/blob/86f1d61adf587e7c8fbcf8289afec6175ceb86de/Civi/Api4/Generic/Traits/DAOActionTrait.php#L167)), so I think we should whitelist instead of blacklist possible actions for this hook, which include e.g. `view` as well, see [here](https://docs.civicrm.org/dev/en/latest/hooks/hook_civicrm_pre/#parameters).

jensschuppe (changes requested):

PR #762, opened 2025-07-01
Fix lock releasing

### LINKED ISSUE #760: Uncaught Error: Call to undefined method Civi\Sepa\Lock\SepaBatchLock::release()
When I perform a rProcess RCUR mandates of the SEPAs (I don't have one) the retry and skip icons appear and the error appears in the civi logs. If I use the skip button continues with the process an finish the proces.
The error:
Error 1: Uncaught Error: Call to undefined method Civi\Sepa\Lock\SepaBatchLock::release() in public_html/wp-content/uploads/civicrm/ext/org.project60.sepa/CRM/Sepa/Logic/Batching.php:255

I modified the file public_html/wp-content/uploads/civicrm/ext/org.project60.sepa/Civi/Sepa/LockSepaBatchLock.php with this code 
  /**
   * Added to provide a public release method for SepaBatchLock,
   * which is expected by Batching.php in CiviSEPA 1.12.1.
   * It delegates the release to the internal CiviCRM lock.
   */
  public function release(): void {
    $this->civiLock->release();
    // Note: If there are scenarios where the asyncLock also needs to be
    // released directly via this public method, that logic would need
    // to be added here. However, based on the error, the primary
    // expectation is for the civiLock part.
  }

inserted in the line 51 to 62 of the file and now not error in the logs and works fine in the process of RCUR mandates groups
Is this fix feasible or will it cause problems? I'm not a programming expert, I'm just trying to help where I can.

Thanks in advance, I hope I'm doing well.
Jon

dontub:
The lock is released on destruction. There's no need to call `release()` at this spot.

The lines were removed in the [PR to master](https://github.com/Project60/org.project60.sepa/pull/744/files#diff-35c2a4e1a3012db2727c48c432df5dd6c2bf9df48c7fa582b091aed42dc5bc45), but not in the [backport to 1.12.x](https://github.com/Project60/org.project60.sepa/pull/746/files#diff-35c2a4e1a3012db2727c48c432df5dd6c2bf9df48c7fa582b091aed42dc5bc45).

Fixes #760

### DIFF
--- a/CRM/Sepa/Logic/Batching.php
+++ b/CRM/Sepa/Logic/Batching.php
@@ -251,8 +251,6 @@ class CRM_Sepa_Logic_Batching {
 
     // step 6: sync calculated group structure with existing (open) groups
     self::syncGroups($mandates_by_nextdate, $existing_groups, $mode, 'RCUR', $rcur_notice, $creditor_id, $offset!==NULL, $offset===0);
-
-    $lock->release();
   }
 
 
@@ -353,8 +351,6 @@ class CRM_Sepa_Logic_Batching {
 
     // step 4: sync calculated group structure with existing (open) groups
     self::syncGroups($calculated_groups, $existing_groups, 'OOFF', 'OOFF', $ooff_notice, $creditor_id, $offset!==NULL, $offset===0);
-
-    $lock->release();
   }
 
 


dontub:
Wrong base branch. Correct is `1.12.x` #763.

PR #763, opened 2025-07-02
Backport: fix lock releasing

bjendres:
The lock is released on destruction. There's no need to call `release()`
at this spot.


bjendres:
@dontub Could you confirm that this is what you had in mind?

dontub:
> @dontub Could you confirm that this is what you had in mind?

Yes, it is.

PR #767, opened 2025-09-02
Do not check Authorization when getting a SepaMandate in SepaMandate pre hook

peetCreative:
…andate

We have a public contribution page to let people register for payed memberships and use sepapp legacy. After the last update of org.project60.sepa this public contribution page breaks because at the changed place the api call is not authenticated. In my Opinion this is not necessary, because  anonymous users need to fill this contribution page.

### DIFF
--- a/CRM/Sepa/BAO/SEPAMandate.php
+++ b/CRM/Sepa/BAO/SEPAMandate.php
@@ -52,12 +52,11 @@ class CRM_Sepa_BAO_SEPAMandate extends CRM_Sepa_DAO_SEPAMandate implements HookI
       }
       else {
         // existing mandate, get creditor
-        $params['creditor_id'] =
-          SepaMandate::get(TRUE)->addSelect('creditor_id')->addWhere(
-            'id',
-            '=',
-            $params['id']
-          )->execute()->single()['creditor_id'];
+        $params['creditor_id'] = SepaMandate::get(FALSE)
+          ->addSelect('creditor_id')
+          ->addWhere('id', '=', $params['id'])
+          ->execute()
+          ->single()['creditor_id'];
       }
     }
     $creditor = SepaCreditor::get(FALSE)


dontub:
Actually that change was introduced by @jensschuppe in [#731](https://github.com/Project60/org.project60.sepa/pull/731/files#diff-cdd059fb691dcddc967eee350a2e629f11063e57c1e4bd793877eff47d10b8e4R55). So I think he can better judge it than me.

peetCreative:
@jensschuppe I agree, noone without permission should be able to edit an existing mandate. But it should be possible to create a mandate with anonymous user. Because we for public contribution pages we need that anyone can create a mandate. Maybe than we should differentiate between create (not checking) and edit (checking).

jensschuppe:
Looking at the code, that's exactly what is supposed to happen: for new mandates, use the default creditor; for existing mandates, load the mandate (with permission check) for retrieving its creditor:

https://github.com/Project60/org.project60.sepa/blob/fdfb35726568875cf32a3a37fa7ee56bae39fe44/CRM/Sepa/BAO/SEPAMandate.php#L48-L61

If this part of the code is being called during creation of new mandates, we'd have to look at a stacktrace.

jensschuppe:
This is indeed an update of an existing mandate, issued from the CiviSEPA payment processor here:
https://github.com/Project60/org.project60.sepapp/blob/34abdee34e5e8485b4c2a39f132679c11bf52527/CRM/Core/Payment/SDD.php#L333

I'm not sure defining `checkPermissions` in this APIv3 call will end up as a parameter in the hook implementation of CiviSEPA where the permission needs to be checked - if so, we could pass that instead of `TRUE`.
This needs debugging, which we can't currently provide, but maybe you can reach out to the maintainers of org.project60.sepapp and let them at least test that …

peetCreative:
Falls ein Kontakt im Formular erkannt wird vom System gematcht  passiert ein ähnlicher Fehler.

Der kann gelöst werden, mittels dieses Commits. https://github.com/Project60/org.project60.sepa/pull/767/commits/e0fe1b4edf602ebd7772458332baf61e73a6873c

Stack Trace folgt

jensschuppe:
APIv3 actions implicitly accept a `checkPermissions` parameter which defaults to `TRUE` for REST calls, and should default to `FALSE` for direct calls via `civicrm_api3()`, but I'm not sure it's actually explicitly set to `FALSE` when it's missing or the implementation just checks for `empty()`.

So I was thinking of explicitly passing it in the `sepapp` extension here: https://github.com/Project60/org.project60.sepapp/blob/34abdee34e5e8485b4c2a39f132679c11bf52527/CRM/Core/Payment/SDD.php#L333 - the assumption is that an explicitly set `checkPermissions` parameter *might* be passed through to *CiviSEPA*'s hook implementation parameters so that the APIv4 call within that hook can evaluate it. But I haven't tested this.

In order to debug that, for a call to `civicrm_api3('SepaMandate', 'create', […])` with an ID and `checkPermissions => FALSE` you could inspect whether the `checkPermissions` parameter ends up here: https://github.com/Project60/org.project60.sepa/blob/2cc02fa626cb6b0ff3e53f0f6078d0db97e0477f/CRM/Sepa/BAO/SEPAMandate.php#L44 and if so, pass it here: https://github.com/Project60/org.project60.sepa/blob/2cc02fa626cb6b0ff3e53f0f6078d0db97e0477f/CRM/Sepa/BAO/SEPAMandate.php#L56 instead of `TRUE`.

peetCreative:
I made another change so now the permission 'make online contribution' is checked and if the user has the permission check permission is set to false.

However, I come around checking if the PaymentProccesor NG is working for me and this payment processor circumvents the problem described here.

jensschuppe:
I had another closer look and came to the conclusion that your initial take on this was the right thing to do, because the pre hook for a sepa mandate should have already passed permission checks. A pre hook is supposed to not abort the action, so it seems ok to just load the same mandate that has already been touched (as it's the pre hook for that very mandate).

Sorry for the confusion along the way, but permission checking is never an easy thing.

For your other change in your last commit, please open a separate PR if that's still an issue.

PR #768, opened 2025-09-17
Update call to deprecated Smarty::get_template_vars

colemanw:
Replaces deprecated function name with the forward-compat equivalent.

Note: the new getTemplateVars function has been available since CiviCRM 5.70.

### DIFF
--- a/info.xml
+++ b/info.xml
@@ -12,8 +12,7 @@
   <version>1.13.0-beta1</version>
   <develStage>beta</develStage>
   <compatibility>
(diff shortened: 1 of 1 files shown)


jensschuppe (changes requested):
Core version compatibility for < `5.70` will have to be removed: https://github.com/Project60/org.project60.sepa/blob/fdfb35726568875cf32a3a37fa7ee56bae39fe44/info.xml#L16

PR #774, opened 2025-10-27
Wordpress links only work by setting htmlize to false

vurt2:
This is a common problem, see also this PR for civioffice: https://github.com/systopia/de.systopia.civioffice/pull/68

Best
Martin


dontub:
Thanks @vurt2!

PR #783, opened 2026-04-02
Replace html spaces with regular spaces

colemanw:
Following a [recent CiviCRM security update](https://civicrm.org/blog/dev-team/civicrm-security-release-6121-610-esr), all field labels are expected to be plain text. Html is escaped. In this case the fix is easy.

Fixes the following problem in CiviCRM 6.12:

image


colemanw:
Thanks @jensschuppe

PR #784, opened 2026-04-21
don't check permission on getLastMandateOfContact

peetCreative:
We use civisepa in combination with civisepapp, to provide contribution pages with direct debit  (german "Lastschrift") payment processors. Filling a Contribution page, creates a contact with a SepaMandate and a according pending contribution. If someone is making a second contribution the contact is matched by the name and the mandate and the contribution is added there.

For the case that a contact is matched civisepa runs into function getLastMandateOfContact (see stacktrace below) and tries to load the last mandate of the contact. Which seems to be not neccesary. But an anonymous user has of course not the right to view mandates. There I get a permission error, and the contribution fails.

grafik

This PR is a quickfix to disable check permission for this api-call.

However the better question is, why in this case the last mandate of the contact is loaded. I think this is not necessary.

I also want to link PR https://github.com/Project60/org.project60.sepa/pull/767 . At the end I thought using sepapp ng payment processor would solve the described problem. This is not anymore the case.

### DIFF
--- a/CRM/Sepa/BAO/SEPAMandate.php
+++ b/CRM/Sepa/BAO/SEPAMandate.php
@@ -896,14 +896,12 @@ class CRM_Sepa_BAO_SEPAMandate extends CRM_Sepa_DAO_SEPAMandate implements HookI
       return FALSE;
     }
 
-    $dao = CRM_Core_DAO::executeQuery(
-      "SELECT * FROM civicrm_sdd_mandate WHERE contact_id = %1 ORDER BY id DESC LIMIT 1",
-      [1 => [$cid, 'Integer']]
-    );
-    if ($dao->fetch()) {
-      return $dao->toArray();
-    }
-    return FALSE;
+    return SepaMandate::get(FALSE)
+        ->addWhere('contact_id', '=', $cid)
+        ->addOrderBy('id', 'DESC')
+        ->setLimit(1)
+        ->execute()
+        ->first() ?? FALSE;
   }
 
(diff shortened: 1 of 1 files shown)


peetCreative:
I understand, that the function fillLastMandateTokenValues is called in the hook sepa_evaluate_tokens. What is this last mandate good for if I actually only want to create a new mandate?

CRM/Sepa/BAO/SEPAMandate.php
```diff
@@ -896,20 +896,14 @@ public static function getLastMandateOfContact($cid) {
...
-      return FALSE;
-    }
+    $dao = CRM_Core_DAO::executeQuery(
+      "SELECT * FROM civicrm_sdd_mandate WHERE contact_id = %1 ORDER BY id DESC LIMIT 1",
+      [1 => [$cid, 'Integer']]
+    );
+    if ($dao->fetch()) {
+      return $dao->toArray();
+    }
+    return FALSE;
```
dontub: I think we should prefer APIv4 calls over SQL. So I'd expect this to work:
```suggestion
    return SepaMandate::get(FALSE)
        ->addWhere('contact_id', '=', $cid)
        ->addOrderBy('id', 'DESC')
        ->setLimit(1)
        ->execute()
        ->first() ?? FALSE;
```

dontub (changes requested):

dontub:
Thanks @peetCreative.

PR #789, opened 2026-04-24
Apply extension template

jensschuppe:
*systopia-reference: 32610*


dontub:
As of now there are mainly code style fixes. Some deprecated code is replaced (e.g. `civicrm_api3()` is used instead of `civicrm_api()`). All methods that had no visibility are now public, though they might have a more restricted visibility.

PR #790, opened 2026-04-29
Fixes issue with log file on standalone

jaapjansma:
On CiviCRM standalone CiviSepa fails with repairing mandates. As the log file is not in `ConfigAndLog` any more but in `private/log`. 

This PR sets CiviSepa to use `[civicrm.log]`  directory.


dontub:
Thanks!

PR #791, opened 2026-04-30
Fix for 'Failed to create OOFF mandate' in the Create Mandate Screen for CiviCRM 12.1

kainuk:
When On creating an OOFF mandate the screen shows the following error.

    Error Failed to create OOFF mandate. Error was: Mandatory key(s) missing from params array: receive_date

The reason is that the collection date is not filled.

The underlying technical reason is that the datapicker works differently in CiviCRM 6. 12.1 as in CiviCRM 5


dontub:
Thanks @kainuk.

PR #802, opened 2026-06-18
Add the IBAN and the mandate reference on the contribution list page

### LINKED ISSUE #801: Show the mandate reference and the IBAN in the contributions list
When clicking on the Contributions button next to an open group or a closed group. It list the contributions, the contact but not the mandate reference and nor the IBAN. It would be useful if this was visible in certain use cases.

jaapjansma:
Fixes #801.

**Background**

At one of our clients we had an issue where the generated contributions ended up with all the same recurring contribution . 

This was not visible in the user interface and only visible the bank file. As each transaction was deducted from the same bank account. (So there was one donor who paid for all donors)

One of the things which would likely prevented this is when the IBAN and the mandate reference would have been visible in the UI.

**Before**

When clicking on *Contributions* next to a SEPA group no info was given on the mandate.

**After**

The mandate reference and the IBAN are visible.

### DIFF
--- a/CRM/Sepa/Page/ListGroup.php
+++ b/CRM/Sepa/Page/ListGroup.php
@@ -109,6 +109,7 @@ class CRM_Sepa_Page_ListGroup extends CRM_Core_Page {
           $contribution['mandate.iban'] = $mandate['iban'];
         }
         catch (\Exception $e) {
+          // @ignoreException
         }
       }
       $contributions[] = [
--- a/CRM/Sepa/Page/ListGroup.php
+++ b/CRM/Sepa/Page/ListGroup.php
@@ -19,6 +19,7 @@ declare(strict_types = 1);
 use CRM_Sepa_ExtensionUtil as E;
 use Civi\Api4\SepaTransactionGroup;
 use Civi\Api4\Contribution;
+use Civi\Api4\SepaMandate;
 
 /**
  * back office sepa group content viewer
@@ -36,6 +37,7 @@ class CRM_Sepa_Page_ListGroup extends CRM_Core_Page {
         ->selectRowCount()
         ->addSelect(
           'id',
+          'type',
           'reference',
           'status_id',
           'COUNT(DISTINCT contribution.id) AS total_count',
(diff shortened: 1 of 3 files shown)


jaapjansma:
The PHP Stan checks are failing but I am not sure why and what can be done about it.

CRM/Sepa/Page/ListGroup.php
```diff
@@ -75,9 +76,25 @@ public function run() {
...
+        ['mandate.entity_id', '=', 'id', TRUE]
+      );
+    }
+    else {
+      // This is a recurring a group.
+      $contributionApi->addJoin('SepaMandate AS mandate', 'INNER', NULL,
+        ['mandate.entity_table', '=', 'civicrm_contribution_recur', FALSE],
+        ['mandate.entity_id', '=', 'contribution_recur_id', TRUE]
+      );
+    }
```
jensschuppe: The `INNER` joins will ignore contributions without a mandate (which can happen when deleting the mandate and keeping the contributions). Let's use a `LEFT` join instead and just not display the IBAN and mandate reference for such contributions.

jensschuppe (changes requested):

CRM/Sepa/Page/ListGroup.php
jaapjansma: I have converted it to a LEFT JOIN

jensschuppe:
Noting that this depends on #803 as otherwise the batching query will return different results.

The tests not failing is actually a sign we should add more tests …

jaapjansma:
@jensschuppe the php unit is failing on the function **data_processor_postInstall** howver in the `dataprocessor.php` file that function is removed a while ago, see: https://lab.civicrm.org/extensions/dataprocessor/-/commit/eef9cd6b91a811edcbff9866e4a95e46ab93ff93?file_path=dataprocessor.civix.php#line_f717dbb74_126

Since then there are two new versions released of the data processor extension. And it seems that the phpunit here is running an old version.

jaapjansma:
@jensschuppe ah now all checks have passed!

jensschuppe:
Needs rebase once #804 is merged (and probably a refactoring given that this currently contains a workaround for the join issue fixed with switching to EFv2).

PR #804, opened 2026-06-18
Convert entity schema definition to Entity Framework v2

### LINKED ISSUE #803: Superfluous join on `first_contribution_id` with API4 when joining with mandate
## Overview

I disovered this issue with the Civi Sepa extension. But it might occur in other places as well.

In the sepa extension we have an entity SepaMandate with the following fields:

- Entity Table (either civicrm_contribution or civicrm_contribution_recur)
- Entity ID
- First Contribution ID (which could be NULL or the ID of the first contribution of a recurring)
- ... (and there more fields)

Now when we do an API4 call on Contribution and joining the Sepa Mandate on the entity table and the entity ID we get another join clause on first contribution ID.

The API call looks like

```php

\Civi\Api4\Contribution::get(TRUE)
  ->addJoin('SepaMandate AS sepa_mandate', 'LEFT', 
    ['sepa_mandate.entity_table', '=', 'civicrm_contribution', FALSE], 
    ['sepa_mandate.entity_id', '=', 'id', TRUE])
  ->execute();
```

## Reproduction steps

1. Install CiviSepa
2. Create a contact
3. Create a ONE-OFF mandate for this contact
4. Go to **Contributions** \> **Sepa Dashboard**  and click on Update ONE-OFF
5. Now run the above API call

## Current behaviour

_The following SQL is generated by the API_

```sql
SELECT
  `a`.`id` AS `id`
FROM `civicrm_contribution` a
LEFT JOIN (`civicrm_sdd_mandate` `sepa_mandate`)
  ON `sepa_mandate`.`first_contribution_id` = `a`.`id`
  AND `sepa_mandate`.`entity_table` = "civicrm_contribution"
  AND `sepa_mandate`.`entity_id` = `a`.`id`
WHERE (`a`.`is_test` = "0")
  AND (`a`.`is_template` = "0")
```

The ON statement contains a join on the first_contribution_id field. Which is wrong.

## Expected behaviour

_I would expect the following sql query_

```sql
SELECT
  `a`.`id` AS `id`
FROM `civicrm_contribution` a
LEFT JOIN (`civicrm_sdd_mandate` `sepa_mandate`)
  ON `sepa_mandate`.`entity_table` = "civicrm_contribution"
  AND `sepa_mandate`.`entity_id` = `a`.`id`
WHERE (`a`.`is_test` = "0")
  AND (`a`.`is_template` = "0")
```

## Environment information

* **Browser:** _Firefox 59.0.1/Chrome 78.0.3904/Safari 13_
* **CiviCRM:** _6.12 (I suspect older and newer versions to have this bug)_

## Comments

It might that this is caused by the sepa extension it self. But I am not sure about it.

See as well: https://lab.civicrm.org/dev/core/-/work_items/6592

jaapjansma:
Fixes #803.

## Overview

I disovered this issue with the Civi Sepa extension. But it might occur in other places as well.

In the sepa extension we have an entity SepaMandate with the following fields:

- Entity Table (either civicrm_contribution or civicrm_contribution_recur)
- Entity ID
- First Contribution ID (which could be NULL or the ID of the first contribution of a recurring)
- ... (and there more fields)

Now when we do an API4 call on Contribution and joining the Sepa Mandate on the entity table and the entity ID we get another join clause on first contribution ID.

The API call looks like

```php

\Civi\Api4\Contribution::get(TRUE)
  ->addJoin('SepaMandate AS sepa_mandate', 'LEFT', 
    ['sepa_mandate.entity_table', '=', 'civicrm_contribution', FALSE], 
    ['sepa_mandate.entity_id', '=', 'id', TRUE])
  ->execute();
```

## Reproduction steps

1. Install CiviSepa
2. Create a contact
3. Create a ONE-OFF mandate for this contact
4. Go to **Contributions** \> **Sepa Dashboard**  and click on Update ONE-OFF
5. Now run the above API call

## Current behaviour

_The following SQL is generated by the API_

```sql
SELECT
  `a`.`id` AS `id`
FROM `civicrm_contribution` a
LEFT JOIN (`civicrm_sdd_mandate` `sepa_mandate`)
  ON `sepa_mandate`.`first_contribution_id` = `a`.`id`
  AND `sepa_mandate`.`entity_table` = "civicrm_contribution"
  AND `sepa_mandate`.`entity_id` = `a`.`id`
WHERE (`a`.`is_test` = "0")
  AND (`a`.`is_template` = "0")
```

The ON statement contains a join on the first_contribution_id field. Which is wrong.

## Expected behaviour

_I would expect the following sql query_

```sql
SELECT
  `a`.`id` AS `id`
FROM `civicrm_contribution` a
LEFT JOIN (`civicrm_sdd_mandate` `sepa_mandate`)
  ON `sepa_mandate`.`entity_table` = "civicrm_contribution"
  AND `sepa_mandate`.`entity_id` = `a`.`id`
WHERE (`a`.`is_test` = "0")
  AND (`a`.`is_template` = "0")
```

## Environment information

* **Browser:** _Firefox 59.0.1/Chrome 78.0.3904/Safari 13_
* **CiviCRM:** _6.12 (I suspect older and newer versions to have this bug)_

## Comments

It might that this is caused by the sepa extension it self. But I am not sure about it.

See as well: https://lab.civicrm.org/dev/core/-/work_items/6592


jaapjansma:
Ideally this should be solved by moving to the new entity schema (v2). However for now this eems easier to fix.

colemanw:
Looks right 👍

CRM/Sepa/DAO/SEPAMandate.php
```diff
@@ -586,6 +593,13 @@ public static function &fieldKeys() {
     return Civi::$statics[__CLASS__]['fieldKeys'];
   }
 
+  public static function entityTableOptions() {
+    return [
+      'civicrm_contribution' => 'civicrm_contribution',
+      'civicrm_contribution_recur' => 'civicrm_contribution_recur'
+    ];
+  }
+
```
jensschuppe: This should live within the BAO class instead of the DAO (the latter gets recreated automatically by civix and will get overwritten on the next civix update).

CRM/Sepa/DAO/SEPAMandate.php
```diff
@@ -586,6 +593,13 @@ public static function &fieldKeys() {
     return Civi::$statics[__CLASS__]['fieldKeys'];
   }
 
+  public static function entityTableOptions() {
+    return [
+      'civicrm_contribution' => 'civicrm_contribution',
+      'civicrm_contribution_recur' => 'civicrm_contribution_recur'
```
jensschuppe: Implementations in Core define translatable entity names as values, so shouldn't this look like the following?
```suggestion
      'civicrm_contribution' => E::ts('Contribution'),
      'civicrm_contribution_recur' => E::ts('Recurring contribution')
```

CRM/Sepa/DAO/SEPAMandate.php
```diff
@@ -304,6 +304,9 @@ public static function &fields() {
           'table_name' => 'civicrm_sdd_mandate',
           'entity' => 'SEPAMandate',
           'bao' => 'CRM_Sepa_DAO_SEPAMandate',
+          'pseudoconstant' => [
+            'callback' => ['CRM_Sepa_DAO_SEPAMandate', 'entityTableOptions']
+          ],
```
jensschuppe: This would have to be defined in the schema XML (since we haven't switched to Entity Framework v2 yet, in which case the definition would belong to the corresponding `*.entityType.php` file). DAO files get overwritten by civix.

CRM/Sepa/DAO/SEPAMandate.php
```diff
@@ -323,6 +326,10 @@ public static function &fields() {
           'table_name' => 'civicrm_sdd_mandate',
           'entity' => 'SEPAMandate',
           'bao' => 'CRM_Sepa_DAO_SEPAMandate',
+          'entity_reference' => [
+            'dynamic_entity' => 'entity_table',
+            'key' => 'id',
+          ],
```
jensschuppe: This would have to be defined in the schema XML (since we haven't switched to Entity Framework v2 yet, in which case the definition would belong to the corresponding `*.entityType.php` file). DAO files get overwritten by civix.

jensschuppe (changes requested):
I'm not sure I understand where the additional join condition derives from. It's being added by guesswork here:
https://github.com/civicrm/civicrm-core/blob/c78228df4bdd9c9f834b528e930d78d9b0774be3/Civi/Api4/Query/Api4SelectQuery.php#L576-L597
@colemanw would you be able to explain what this is supposed to do and why this is adding join conditions when the given join is perfectly fine without it?

Also, @jaapjansma, with your changes applied, the query doesn't look correct either:
```sql
SELECT
  `a`.`id` AS `id`
FROM `civicrm_contribution` a
LEFT JOIN (`civicrm_sdd_mandate` `sepa_mandate`)
  ON (`sepa_mandate`.`entity_table` IS NULL OR ((`sepa_mandate`.`entity_table` = 'civicrm_contribution' OR `sepa_mandate`.`entity_table` = 'civicrm_contribution_recur')))
  AND `sepa_mandate`.`entity_table` = \"civicrm_contribution\"
  AND `sepa_mandate`.`entity_id` = `a`.`id`
WHERE (`a`.`is_test` = \"0\")
  AND (`a`.`is_template` = \"0\")
LIMIT 25
OFFSET 0
```
corresponding API call:
```php
\Civi\Api4\Contribution::get(TRUE)
  ->addJoin(
    'SepaMandate AS sepa_mandate',
    'LEFT',
    ['sepa_mandate.entity_table', '=', '"civicrm_contribution"'],
    ['sepa_mandate.entity_id', '=', 'id']
  )
  ->execute();
```
Notice the excess join conditions for all `entity_table` options (including `NULL`) when only `"civicrm_contribution"` was requested.

jaapjansma:
I am going to convert it to entity v2  and I discover that the field entity_table was missing a required flag.

CRM/Sepa/DAO/SEPAMandate.php
jaapjansma: This is moved to the BAO

CRM/Sepa/DAO/SEPAMandate.php
jaapjansma: This is moved to *.entityType.php file.

CRM/Sepa/DAO/SEPAMandate.php
jaapjansma: Yes you are right. I have updated this.

CRM/Sepa/DAO/SEPAMandate.php
jaapjansma: This is moved to the BAO

jaapjansma:
I have updated the PR with a move to entity v2. And then added the missing pseudoconstant on entity table and made entity table required.

jensschuppe:
Can you also have a look whether this part of the code is covered by tests (the batching of transaction groups) - I have a feeling we should add some more explicit coverage for changes like this (actually not touching CiviSEPA batching logic, but the queries being so different has me worrying).

jaapjansma:
@jensschuppe I am not sure why phpstan is failing. Can you have a look at that? As it is failing over changes I did not make.

jaapjansma:
> Can you also have a look whether this part of the code is covered by tests (the batching of transaction groups) - I have a feeling we should add some more explicit coverage for changes like this (actually not touching CiviSEPA batching logic, but the queries being so different has me worrying).

I did have a look at the unit tests those are very basic. There is one for mandate and a RCUR batch generation. It looks like test coverage could be improved a lot. But I am not sure whether my client and is willing to fund that.

CRM/Sepa/DAO/SEPATransactionGroup.php
```diff
@@ -1,457 +1,30 @@
...
-use CRM_Sepa_ExtensionUtil as E;
-
-/**
- * Database access object for the SEPATransactionGroup entity.
+ * DAOs are a primary source for metadata in older versions of CiviCRM (<5.74)
+ * and are required for some subsystems (such as APIv3).
+ *
+ * This stub provides compatibility. It is not intended to be modified in a
+ * substantive way. Property annotations may be added, but are not required.
+ * @property int $id
```
jensschuppe: I'm a bit in doubt about the type here, because I suppose all the properties annotations have been created automatically by civix and this came out as `string` - however, this will always be an integer, and it's also an integer in DB. Are these supposed to be adjusted manually? If so, all the other attributes (in all CiviSEPA entities) might need checking for the correct type.

jaapjansma: Strange. I have not seen those properties before. Looks like a an error made by  civix

jaapjansma: Usually those property definitions are not there

jensschuppe: These annotations are required because the properties are implicitly declared in the base class, but the type hints seem off.

jaapjansma: I have disovered the issue I did not use the latest civix to generate those.

jaapjansma: I have updated the PR with a new version of civix biolerplate

colemanw: Now phpstan isn't happy with the *lack* of property annotations. 😭

jaapjansma:
I have checked the failures of the PHP Unit tests. But those are failing because of a mis setup of php unit? For example it failing because the method `copyValues` does not exist on the DOA classes. But it does clearly exists in core. So I am a bit clueless on how to solve this.

@jensschuppe can you help me with this?

colemanw:
> it failing because the method copyValues does not exist

It should exist because the class is ultimately inherited from CRM_Core_DAO (via `class_alias('CRM_Core_DAO_Base', 'CRM_Sepa_DAO_Base');`)

CRM/Sepa/DAO/SEPATransactionGroup.php
jensschuppe: PHPStan doesn't consider class aliases without explicit definition in PHPStan's bootstrap. Is there actually a reason for using an aliased class `CRM_Sepa_DAO_Base` instead of `CRM_Core_DAO_Base`, @colemanw? This would make it easier to ignore the errors in PHPStan's configuration.

colemanw: @jensschuppe yes it would be fine to change these DAO stubs to inherit directly from `CRM_Core_DAO_Base`. The class alias was added by @totten for backward-compat but shouldn't be needed if the info.xml specifies a reasonably recent CiviCRM-core version.

jensschuppe:
@jaapjansma Seems CiviSEPA's soft dependencies need updates (see PHPUnit errors regarding covariance of a class in dataprocessor). Can you look into that?

I know this feels cumbersome but I'd really like the tests to pass on this one and compare the results, as this touches a core part of CiviSEPA and fixing errors involving actual transactions has always been a mess in the past. Also, we all profit from strict quality standards, don't we? :wink:

jaapjansma:
@jensschuppe I agree the test should be passing. What I do not understand is why phpstand is failing on installing the data processor extension. As that extension is not a hard requirement.

I did find this line: https://github.com/Project60/org.project60.sepa/blob/d543a366df062bb429fdf3c2a854e46f428bf62b/phpstan.neon.dist#L30 which seems to instruct phpstan to install the data processor extension
It could also be this line:
https://github.com/Project60/org.project60.sepa/blob/d543a366df062bb429fdf3c2a854e46f428bf62b/.github/workflows/phpstan.yml#L86

jaapjansma:
@jensschuppe I just tried to install the data processor extension in a clean civicrm environment and the data processor gave an error.

jaapjansma:
@jensschuppe the errors in the data processor extension are resolved. Can we restart the tests again?

jensschuppe:
@jaapjansma looking at the [diff](https://lab.civicrm.org/extensions/dataprocessor/-/compare/1.151...1.152) it seems the issue has not been resolved in dataprocessor yet. The issue would be solved by adding the `:void` return type hint to the affected method in dataprocessor. See this [example commit](https://github.com/systopia/de.systopia.identitytracker/pull/31/changes/7e061b3a3c97bea0d375d136c38462c9bac98840) in another extension.

jaapjansma:
@jensschuppe we now have another error:

**Class "Civi\ActionProvider\Action\AbstractAction" not found**

I do not understand why phpunit complains about this as it downloads the action provider extension.

jensschuppe:
@jaapjansma one issue remaining in dataprocessor with current Symfony:
`Declaration of Civi\DataProcessor\Output\AbstractApi::getSubscribedEvents() must be compatible with Symfony\Component\EventDispatcher\EventSubscriberInterface::getSubscribedEvents(): array in /home/runner/work/org.project60.sepa/civicrm-standalone/ext/dataprocessor/Civi/DataProcessor/Output/AbstractApi.php on line 84`

> Class "Civi\ActionProvider\Action\AbstractAction" not found
> 
> I do not understand why phpunit complains about this as it downloads the action provider extension.

We can't use the `scan-classes` mixin here as there are optional dependencies (like *action-provider*)

With the tests passing, those concerns are still valid: https://github.com/Project60/org.project60.sepa/pull/804#issuecomment-4751392325 and https://github.com/Project60/org.project60.sepa/pull/804#pullrequestreview-4531625274

Thanks for your patience though with the tools!

jaapjansma:
@jensschuppe so we have managed to pass the phpstan and phpunittests.

jensschuppe:
> With the tests passing, those concerns are still valid: [#804 (comment)](https://github.com/Project60/org.project60.sepa/pull/804#issuecomment-4751392325) and [#804 (review)](https://github.com/Project60/org.project60.sepa/pull/804#pullrequestreview-4531625274)

@jaapjansma Can you do that comparison of actual SQL queries before and after this change? There seems to be an unnecessary condition … this would really help with the review.

Also, the branch needs rebasing onto the current `master`.

jaapjansma:
@jensschuppe below the query generated with this patch. I do not see any strange things (anymore) in this query.

I did the following api call:

```php
\Civi\Api4\Contribution::get(TRUE)
      ->addSelect('id')
      ->addJoin('SepaMandate AS sepa_mandate', 'LEFT', 
      ['sepa_mandate.entity_table', '=', 'civicrm_contribution', FALSE], 
      ['sepa_mandate.entity_id', '=', 'id', TRUE])
->execute();
```

Resulting in this query


```sql
SELECT `a`.`id` AS `id`
FROM `civicrm_contribution` a 
LEFT JOIN (`civicrm_sdd_mandate` `sepa_mandate`) ON 
  ((`sepa_mandate`.`entity_table` = 'civicrm_contribution' OR `sepa_mandate`.`entity_table` = 'civicrm_contribution_recur'))
  AND `sepa_mandate`.`entity_table` = "civicrm_contribution" 
  AND `sepa_mandate`.`entity_id` = `a`.`id` 
WHERE (`a`.`is_test` = "0") AND (`a`.`is_template` = "0") 
```

The JOIN condition might seem a little weird. But basically its is generated because `entity_table` and `entity_id` are dynamics keys.

The condition is exists of the following:
* Check whether entity_table is either `civicrm_contribution` or `civicrm_contribution_recur` any other value is not valid
* then match only the ones with entity_table is `civicrm_contribution`
* and then match it on the entity_id column

jaapjansma:
@jensschuppe I have rebases this PR.

jensschuppe:
Rebased onto current `master`. Pending:
* Add EFv2 definition for new `SepaMandateLink` (`civix convert-entity`)
* Delete XML schema files of entities now defined using EFv2
* Delete `sepa.sql` and stop calling them in the upgrader
* Check SQL table descriptions (mismatch XML vs SQL definition, and a random `FIXME`)

We'll try to do that and merge this PR for the upcoming release.

sepa.php
```diff
@@ -346,36 +341,7 @@ function sepa_civicrm_post($op, $objectName, $objectId, &$objectRef): void {
...
-    'name' => 'SepaSddFile',
-    'class' => 'CRM_Sepa_DAO_SEPASddFile',
-    'table' => 'civicrm_sdd_file',
-  ];
-  $entityTypes[] = [
-    'name' => 'SepaContributionGroup',
-    'class' => 'CRM_Sepa_DAO_SEPAContributionGroup',
-    'table' => 'civicrm_sdd_contribution_txgroup',
-  ];
   $entityTypes[] = [
```
dontub: I think this is obsolete, too.

schema/SEPAContributionGroup.entityType.php
```diff
@@ -0,0 +1,56 @@
...
+  'class' => 'CRM_Sepa_DAO_SEPAContributionGroup',
+  'getInfo' => fn() => [
+    'title' => E::ts('SEPAContribution Group'),
+    'title_plural' => E::ts('SEPAContribution Groups'),
+    'description' => E::ts('Link Contributions to TX Group'),
+  ],
+  'getIndices' => fn() => [
+    'contriblookup' => [
+      'fields' => [
+        'contribution_id' => TRUE,
```
dontub, line 16: This was marked as unique in `sepa.sql`.

schema/SEPAContributionGroup.entityType.php
```diff
@@ -0,0 +1,56 @@
...
+      ],
+    ],
+    'txglookup' => [
+      'fields' => [
+        'txgroup_id' => TRUE,
```
dontub: There should already be an FK constraint making this additional index superfluous. ~~The existing key could be deleted in an update step.~~

Edit: There's no need to delete the index in an update step. It was just explicitly named. It would have been created anyways because of the FK constraint. (Deleting it is actually impossible because it's used by the FK constraint.)

schema/SEPACreditor.entityType.php
```diff
@@ -0,0 +1,179 @@
...
+
+return [
+  'name' => 'SepaCreditor',
+  'table' => 'civicrm_sdd_creditor',
+  'class' => 'CRM_Sepa_DAO_SEPACreditor',
+  'getInfo' => fn() => [
+    'title' => E::ts('SEPACreditor'),
+    'title_plural' => E::ts('SEPACreditors'),
+    'log' => TRUE,
+    'add' => '4.3',
```
dontub: Actually the `add` attributes could be deleted.

schema/SEPAMandate.entityType.php
```diff
@@ -0,0 +1,174 @@
...
+  'class' => 'CRM_Sepa_DAO_SEPAMandate',
+  'getInfo' => fn() => [
+    'title' => E::ts('SEPAMandate'),
+    'title_plural' => E::ts('SEPAMandates'),
+    'log' => TRUE,
+  ],
+  'getIndices' => fn() => [
+    'reference' => [
+      'fields' => [
+        'reference' => TRUE,
```
dontub, line 16: This was unique in `sepa.sql`.

schema/SEPASddFile.entityType.php
```diff
@@ -0,0 +1,93 @@
...
+        'key' => 'id',
+        'on_delete' => 'SET NULL',
+      ],
+    ],
+    'status_id' => [
+      'title' => E::ts('Status ID'),
+      'sql_type' => 'int unsigned',
+      'input_type' => 'Number',
+      'required' => TRUE,
+      'description' => E::ts('fk to Batch Status options in civicrm_option_values'),
```
dontub, line 78: Maybe add
```php
      'pseudoconstant' => [
        'option_group_name' => 'batch_status',
      ],
```

schema/SEPATransactionGroup.entityType.php
```diff
@@ -0,0 +1,112 @@
...
+  'getIndices' => fn() => [
+    'UI_reference' => [
+      'fields' => [
+        'reference' => TRUE,
+      ],
+      'unique' => TRUE,
+    ],
+    'creditor_id' => [
+      'fields' => [
+        'sdd_creditor_id' => TRUE,
```
dontub: Superfluous because of FK constraint.

schema/SEPATransactionGroup.entityType.php
```diff
@@ -0,0 +1,112 @@
...
+      ],
+      'add' => '4.3',
+    ],
+    'file_id' => [
+      'fields' => [
+        'sdd_file_id' => TRUE,
```
dontub: Superfluous because of FK constraint.

schema/SepaMandateLink.entityType.php
```diff
@@ -0,0 +1,125 @@
...
+    ],
+  ],
+  'getFields' => fn() => [
+    'id' => [
+      'title' => E::ts('ID'),
+      'sql_type' => 'int unsigned',
+      'input_type' => 'Number',
+      'required' => TRUE,
+      'description' => E::ts('Unique SepaMandateLink ID'),
+      'entity_reference' => [
```
dontub: Should be at field `mandate_id`. (Was already incorrect in `MandateLink.xml`.)

dontub (changes requested):

.github/workflows/phpunit.yml
```diff
@@ -123,6 +123,9 @@ jobs:
           cv vars:fill --quiet
           sed -i "s#\"TEST_DB_DSN\": \".*\"#\"TEST_DB_DSN\": \"$DATABASE_URL\"#g" "$CV_CONFIG"
 
+          cv ext:download "action-provider@https://lab.civicrm.org/extensions/action-provider/-/archive/1.195/action-provider-1.195.zip"
+          cv ext:download "dataprocessor@https://lab.civicrm.org/extensions/dataprocessor/-/archive/1.155/dataprocessor-1.155.zip"
```
dontub: Wondering if this is actually necessary because this PR doesn't contain additional tests and the tests already run without these extensions.

jensschuppe: Removed them. Actually a sign for missing tests on the integrations.

dontub (approved):
I think this should be squashed before merge.

colemanw:
Nice work everyone

PR #808, opened 2026-07-06
Migrate APIv3 action SepaMandate.createfull to APIv4

dontub:
The action is now named `createFull` (with uppercase "F").

The APIv3 action is kept, but calls the APIv4 action.

I've found three issues in the previous APIv3 action that are changed in the APIv4 action:

* The type in the call of `\CRM_Sepa_Logic_PaymentInstruments::getPaymentInstrumentsForCreditor()` [here](https://github.com/Project60/org.project60.sepa/blob/a9cdfbb0a4e61f369ee2f8e67e7e1fd4689b5e4b/api/v3/SepaMandate.php#L112) (`$rcur_pi_status`) was either `'OOFF'` or `'RCUR'`, but never `'FRST'`. Though (to my understanding) it should be `'FRST'` for RCUR mandates in status `'FRST'`.
* `contribution_contact_id` overwrites `contact_id` (for `SepaMandate`), but was meant to be used only for the contribution. Reason is the assignment `$create_mandate = $create_contribution;` [here](https://github.com/Project60/org.project60.sepa/blob/a9cdfbb0a4e61f369ee2f8e67e7e1fd4689b5e4b/api/v3/SepaMandate.php#L225) that should have been `$create_mandate = $params;`.
* The field `is_pay_later` is set [here](https://github.com/Project60/org.project60.sepa/blob/a9cdfbb0a4e61f369ee2f8e67e7e1fd4689b5e4b/api/v3/SepaMandate.php#L194) when a `ContributionRecur` is created. Though only `Contribution` has this field, but not `ContributionRecur`. This field isn't set in the APIv4 action anymore. Should it be set for `Contribution`, though?

systopia-reference: 32610

### DIFF
--- a/Civi/Sepa/Api4/Action/SepaMandate/CreateFullAction.php
+++ b/Civi/Sepa/Api4/Action/SepaMandate/CreateFullAction.php
@@ -1,0 +1,171 @@
+<?php
+/*
+ * Copyright (C) 2026 SYSTOPIA GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU Affero General Public License as published by the Free
+ * Software Foundation, either version 3 of the License, or (at your option) any
+ * later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+declare(strict_types = 1);
+
+namespace Civi\Sepa\Api4\Action\SepaMandate;
+
+use Civi\Api4\Contribution;
+use Civi\Api4\ContributionRecur;
+use Civi\Api4\Generic\AbstractCreateAction;
+use Civi\Api4\Generic\Result;
+use Civi\Api4\SepaCreditor;
+use Civi\Api4\SepaMandate;
+use Civi\Sepa\Contribution\PaymentInstrumentDeterminer;
+use Civi\Sepa\Mandate\MandateStatusDeterminer;
+use Civi\Sepa\Util\ContributionUtil;
+
+final class CreateFullAction extends AbstractCreateAction {
+
+  public function _run(Result $result): void {
+    $transaction = \CRM_Core_Transaction::create();
+    $transaction->run(fn() => $this->doRun($result));
+  }
+
+  /**
+   * @throws \CRM_Core_Exception
+   */
+  // phpcs:ignore Generic.Metrics.CyclomaticComplexity.TooHigh
+  public function doRun(Result $result): void {
+    $this->validateValues();
+    // TODO: more sanity checks?
+
+    /**
+     * @var array{
+     *    contact_id: int|numeric-string,
+     *    iban: string,
+     *    creditor_id?: int|numeric-string,
+     *    type: string,
+     *    status?: string,
+     *    payment_instrument_id?: int|numeric-string,
+     *    financial_type_id: int|numeric-string,
+     *    bic?: string,
+     *    contribution_contact_id?: int|numeric-string,
+     *    currency?: string,
+     *    contribution_status_id?: int|numeric-string,
+     *    amount: int|float|numeric-string,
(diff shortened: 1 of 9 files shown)


Civi/Sepa/Api4/Action/SepaMandate/CreateFullAction.php
jensschuppe, line 1: How about moving most of the code into a class independent from the API layer?

Civi/Sepa/Api4/Action/SepaMandate/CreateFullAction.php
```diff
@@ -0,0 +1,200 @@
...
+      $contributionEntity = 'Contribution';
+      $contributionTable  = 'civicrm_contribution';
+      $contributionValues['total_amount'] = $contributionValues['amount'];
+    }
+    else {
+      throw new \CRM_Core_Exception('Unknown mandate type: ' . $values['type']);
+    }
+
+    // create the contribution
+    $contribution = civicrm_api4($contributionEntity, 'create', [
```
jensschuppe: I'd suggest to explicitly create the respective entities calling `Contribution::create()` and `ContributionRecur::create()` for making those calls more easily identifiable. As there are only those two cases, and there is no more common between the two condition branches and the creation, the use of `civicrm_api4()` doesn't seem reasonable to me.

Civi/Sepa/Api4/Action/SepaMandate/CreateFullAction.php
```diff
@@ -0,0 +1,200 @@
...
+    /** @var array<string, mixed> $mandate */
+    $mandate = SepaMandate::create($this->getCheckPermissions())
+      ->setValues($mandateValues)
+      ->execute()
+      ->single();
+
+    $result[] = $mandate;
+  }
+
+  private function determineMandateStatus(string $type, ?int $first_contribution_id): string {
```
jensschuppe: Those helper methods could be moved to a dedicated API-independent class, along with much of the code inside the action itself.

tests/phpunit/CRM/Sepa/BugReproductionTest.php
```diff
@@ -84,22 +84,6 @@ public function testBug629() {
     $mandateDate = 'now - 65 days';
     $monthly_mandate = $this->createMandate(['type' => self::MANDATE_TYPE_RCUR], $mandateDate);
 
-    // NOW mess up the recurring contribution in a way that likely caused #629:
-    //  give the recurring contribution the FRST payment instrument
```
jensschuppe, line 88: This comment kind of contradicts with your assumption that `ContributionRecur` entities should have the `FRST` payment instrument when the corresponding mandate is in `FRST` status. Reading all the issues and PRs that tried to fix errors in this regard, I'm actually not sure what the correct combination of mandate status and payment insturment for `FRST` recurring mandates would be.

Anyway, why is this part of the test being removed?

jensschuppe (commented):
> * The field `is_pay_later` is set [here](https://github.com/Project60/org.project60.sepa/blob/a9cdfbb0a4e61f369ee2f8e67e7e1fd4689b5e4b/api/v3/SepaMandate.php#L194) when a `ContributionRecur` is created. Though only `Contribution` has this field, but not `ContributionRecur`. This field isn't set in the APIv4 action anymore. Should it be set for `Contribution`, though?

If the field doesn't exist on the `ContributionRecur` entity, not passing it doesn't make a difference. If there was code relying on the field being implicitly set for individual `Contribution`s, this wouldn't have worked for quite some time now anyway.

Civi/Sepa/Api4/Action/SepaMandate/CreateFullAction.php
dontub: I've know moved part of the code in separate classes.

Civi/Sepa/Api4/Action/SepaMandate/CreateFullAction.php
dontub: Changed.

tests/phpunit/CRM/Sepa/BugReproductionTest.php
dontub: > Anyway, why is this part of the test being removed?

It had become obsolete because the recurring contribution already had the FRST payment instrument. Though I've reverted it now. It seems that the recurring contribution should always have a RCUR payment instrument and the first normal contribution will get an FRST payment instrument. Later normal contributions will get the payment instrument from the recurring contribution.

jensschuppe (approved):

PR #809, opened 2026-07-09
Allow to suspend and reinstate mandates

dontub:
This adds the possibility to suspend and reinstate mandates.

The code in `\CRM_Sepa_Logic_Batching::updateRCUR()` expects `payment_instrument_id` and `financial_type_id` of the recurring contribution to be *not* `NULL`. Now with the type hints set in accordance to `nullable` as reported by `ContributionRecur.getFields` phpstan reports errors. When the recurring contributions are created, the values are not `NULL`. So if those values weren't manually changed they actually cannot be `NULL`. Shall we add additional where conditions to be sure we don't get `NULL` or throw an exception if one of those values is `NULL`?

~~Documentation is left to do.~~

systopia-reference: 32610


dontub:
The test with CiviCRM 6.10 (`prefer-lowest`) currently fail because payment instruments are expected to be connected to an financial account, see https://github.com/civicrm/civicrm-core/pull/34813. Should we connect the SEPA payment instruments with the financial account with `financial_account_type_id:name = 'Asset'` and `is_default = TRUE` on extension install? Or increase the minimum supported version to 6.13 where there is a fallback?

managed/CustomGroup_civi_sepa_contribution.mgd.php
```diff
@@ -0,0 +1,52 @@
...
+return [
+  [
+    'name' => 'CustomGroup_civi_sepa_contribution',
+    'entity' => 'CustomGroup',
+    'cleanup' => 'unused',
+    'update' => 'unmodified',
+    'params' => [
+      'version' => 4,
+      'values' => [
+        'name' => 'civi_sepa_contribution',
```
jensschuppe: Let's just use the `file` key of the extension which is just `sepa` without the `civi_` prefix.

managed/CustomGroup_civi_sepa_contribution_recur.mgd.php
```diff
@@ -0,0 +1,54 @@
...
+return [
+  [
+    'name' => 'CustomGroup_civi_sepa_contribution_recur',
+    'entity' => 'CustomGroup',
+    'cleanup' => 'unused',
+    'update' => 'unmodified',
+    'params' => [
+      'version' => 4,
+      'values' => [
+        'name' => 'civi_sepa_contribution_recur',
```
jensschuppe: Let's just use the `file` key of the extension which is just `sepa` without the `civi_` prefix.

CRM/Sepa/Logic/Batching.php
```diff
@@ -126,192 +158,161 @@ public static function updateRCUR(
...
-    // RCUR-STEP 4: create the missing contributions, store all in $mandate['mandate_entity_id']
-    foreach ($mandates_by_nextdate as $collection_date => $financial_type_mandates) {
-      foreach ($financial_type_mandates as $financial_type => $mandates) {
-        foreach ($mandates as $index => $mandate) {
-          $recur_id = $mandate['mandate_entity_id'];
-          if (isset($existing_contributions_by_recur_id[$recur_id])) {
+    // RCUR-STEP 3: find already created contributions
+    $existingContributionsByRecurId = [];
+    foreach ($rcontribIdsByCollectionDate as $collectionDate => $rcontribIds) {
+      $existingContributionsByRecurId += Contribution::get(FALSE)
```
jensschuppe, line 218: Did you compare the actual SQL queries being issued with the *financialacls* extension installed? Passing `FALSE` for the `$checkPermissions` parameter would skip the check for ACLs using that extension, while the previous code (a raw SQL query) would not, as *financialacls* monitors both, API calls and SQL queries. So I suppose `Contribution::get(FALSE)` might include contributions the user doesn't have permission for based on financial type.

managed/CustomGroup_sepa_contribution.mgd.php
```diff
@@ -0,0 +1,52 @@
...
+    'entity' => 'CustomGroup',
+    'cleanup' => 'unused',
+    'update' => 'unmodified',
+    'params' => [
+      'version' => 4,
+      'values' => [
+        'name' => 'civi_sepa_contribution',
+        'table_name' => 'civicrm_value_civi_sepa_contribution',
+        'title' => E::ts('CiviSEPA Contribution Fields'),
+        'extends' => 'Contribution',
```
jensschuppe, line 16: The `Contribution` entity needs to be added to the `cg_extend_objects` option group in order to e extended by custom groups.

managed/CustomGroup_sepa_contribution_recur.mgd.php
```diff
@@ -0,0 +1,54 @@
...
+    'entity' => 'CustomGroup',
+    'cleanup' => 'unused',
+    'update' => 'unmodified',
+    'params' => [
+      'version' => 4,
+      'values' => [
+        'name' => 'civi_sepa_contribution_recur',
+        'table_name' => 'civicrm_value_civi_sepa_contribution_recur',
+        'title' => E::ts('CiviSEPA ContributionRecur Fields'),
+        'extends' => 'ContributionRecur',
```
jensschuppe, line 16: The `ContributionRecur` entity needs to be added to the `cg_extend_objects` option group in order to e extended by custom groups.

tests/phpunit/Civi/Sepa/BatchingOnHoldTest.php
```diff
@@ -0,0 +1,271 @@
...
+      ->addValue('status', 'ONHOLD')
+      ->addWhere('id', '=', $mandate['id'])
+      ->execute()
+      ->single();
+
+    ContributionRecur::update(FALSE)
+      ->addValue('civi_sepa_contribution_recur.is_on_hold', TRUE)
+      ->addWhere('id', '=', $mandate['entity_id'])
+      ->execute()
+      ->single();
```
jensschuppe: Why not use `SepaMandate::suspend()` here?

tests/phpunit/Civi/Sepa/BatchingOnHoldTest.php
```diff
@@ -0,0 +1,271 @@
...
+
+    \CRM_Sepa_Logic_Batching::updateRCUR($creditorId, 'FRST');
+
+    // RCUR mandate in ONHOLD should not handled.
```
jensschuppe: Typo in comment.

jensschuppe (changes requested):

managed/CustomGroup_sepa_contribution.mgd.php
dontub: `Contribution` belongs to the entities where CiviCRM supports custom fields out of the box (like for `Contact`).

managed/CustomGroup_sepa_contribution_recur.mgd.php
dontub: `ContributionRecur` belongs to the entities where CiviCRM supports custom fields out of the box (like for `Contact`).

jensschuppe: I thought they'd be in that option group then - seems that just applies for *additional* entity types - nevermind.

managed/CustomGroup_sepa_contribution.mgd.php
jensschuppe: I thought they'd be in that option group then - seems that just applies for *additional* entity types - nevermind.

CRM/Sepa/Logic/Batching.php
dontub: > *financialacls* monitors both, API calls and SQL queries

Didn't know that. However, `->addWhere('contribution_recur_id', 'IN', $rcontribIds)` already limits the returned contributions. So I don't think *financialacls* has to get active here.

Interestingly the custom field is missing in the phpunit tests when `$checkPermissions` is `TRUE`.

tests/phpunit/Civi/Sepa/BatchingOnHoldTest.php
dontub: Changed.

tests/phpunit/Civi/Sepa/BatchingOnHoldTest.php
dontub: Fixed.

managed/CustomGroup_civi_sepa_contribution.mgd.php
dontub: Changed.

managed/CustomGroup_civi_sepa_contribution_recur.mgd.php
dontub: Changed.

CRM/Sepa/Logic/Batching.php
jensschuppe: The individual contributions could still have financial types the user does not have permission for via *financialacls* - maybe we should add a test for that scenario …?

dontub: As discussed there's no need to apply financialacls here because we don't want to replace existing contributions.

dontub:
Regarding nullability of `payment_instrument_id` and `financial_type_id` https://github.com/Project60/org.project60.sepa/pull/809#issue-4846667374: I've changed the type hint as discussed and added a comment.

Regarding https://github.com/Project60/org.project60.sepa/pull/809#issuecomment-4926749856: I've increased the minimal supported CiviCRM version to 6.13 as discussed.

jensschuppe (approved):

PR #813, opened 2026-07-17
Add entity schema for SepaMandateLink (civicrm_sdd_entity_mandate)

dontub:
The entity schema for `SepaMandateLink` was missing and the DAO class was generated by hand. Now the entity is defined in an XML file as the other entities.

An unnecessary index is deleted in an update step.

The entity schema generated with `civix generate:entity-boilerplate` (`auto_install.sql`) for `civicrm_sdd_entity_mandate` is equal to the one now in `sepa.sql`.

Previously the comments in `sepa.sql` weren't completely matching the DAO class.

systopia-reference: 32611


jensschuppe (commented):
Not using *Entity Framework v2* to avoid mixing with legacy schema definitions for the other entities …?

dontub:
> Not using _Entity Framework v2_ to avoid mixing with legacy schema definitions for the other entities …?

Yes, I think all entities should be switched at once using `civix`.

jensschuppe (approved):

jensschuppe:
This causes an error in the upgrader task with MySQL:
```
DROP INDEX IF EXISTS mandate_id ON civicrm_sdd_entity_mandate [nativecode=1064 ** You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'IF EXISTS mandate_id ON civicrm_sdd_entity_mandate' at line 1]
```
Seems `IF EXISTS` is only available for `DROP INDEX` on MariaDB.

PR #820, opened 2026-08-03
Fixes styling for civi sepa dashboard

### LINKED ISSUE #816: Dashboard Colours not working on Walbrook
The colours for the table backgrounds are overwritten by the new CiviCRM theme. Update the styling rule such that it takes precedence over the theme's styling.

Image

knottheory:
Added a stylesheet to overwrite the background colour of the civi sepa dashboard. The colour now use the colour danger, warning, info and success colours by the theme and use the original colours as a fallback if no walbrook theme is installed.

Changed the buttons to use the civicrm template from civicrm to align their styling with the rest of Civi. Added styling for the order of those buttons. Added icons to those buttons.

Those changes fix issue #816.


css/sepa.css
```diff
@@ -0,0 +1,24 @@
+.sepa_dashboard str.submit_missed td {
```
jensschuppe, line 1: Typo, should be `tr.submit_missed`.

css/sepa.css
```diff
@@ -0,0 +1,24 @@
...
+.sepa_dashboard tr.submit_later td {
+  background-color: var(--crm-success-light-color, #008300AA);
+}
+
+.sepa_dashboard tr.submit_closed td {
+  background-color: var(--crm-layer2-bg-color, #00830033);
+}
+
+.sepa_dashboard .sepa_actions {
+  display: flex;
```
jensschuppe, line 22: Why does the table cell have to be `display: flex;`? - this actually causes the buttons to grow in height and shouldn't be necessary.

The buttons are `inline-block` so `text-align: right;` should be enough.

css/sepa.css
```diff
@@ -0,0 +1,24 @@
...
+
+.sepa_dashboard tr.submit_later td {
+  background-color: var(--crm-success-light-color, #008300AA);
+}
+
+.sepa_dashboard tr.submit_closed td {
+  background-color: var(--crm-layer2-bg-color, #00830033);
+}
+
+.sepa_dashboard .sepa_actions {
```
jensschuppe, line 21: Let's use BEM conventions for selectors, making this `.sepa-dashboard__sepa-actions`.

templates/CRM/Sepa/Page/DashBoard.tpl
```diff
@@ -70,17 +70,19 @@
...
+<table class="sepa_dashboard" id="options">
+  <thead>
+    <tr role="row">
+      <th class="sorting" aria-controls="sepa-option">{ts domain="org.project60.sepa"}Group Name{/ts}</th>
+      <th class="sorting" aria-controls="sepa-option">{ts domain="org.project60.sepa"}Status{/ts}</th>
+      <th class="sorting" aria-controls="sepa-option">{ts domain="org.project60.sepa"}Type{/ts}</th>
+      <th class="sorting" aria-controls="sepa-option">{ts domain="org.project60.sepa"}Submission{/ts}</th>
+      <th class="sorting" aria-controls="sepa-option">{ts domain="org.project60.sepa"}Collection{/ts}</th>
+      <th class="sorting" aria-controls="sepa-option">{ts domain="org.project60.sepa"}Transactions{/ts}</th>
+      <th class="sorting" aria-controls="sepa-option">{ts domain="org.project60.sepa"}Total{/ts}</th>
```
jensschuppe, line 82: Let's take the chance getting rid of the `domain` attribute for translations by wrapping the entire template in `{crmScope extensionKey='org.project60.sepa'}…{/crmScope}` tags (applies to all occurences of `{ts}` in this template).

templates/CRM/Sepa/Page/DashBoard.tpl
```diff
@@ -105,27 +107,26 @@
       {$group.nb_contrib}
     </td>
     <td style="white-space:nowrap;">{$group.total|crmMoney:$group.currency}</td>
-    <td>
-      <a href="{crmURL p="civicrm/sepa/listgroup" q="group_id=$group_id"}" class="button button_view">{ts domain="org.project60.sepa"}Contributions{/ts}</a>
+    <td class="sepa_actions">
+      {crmButton href="{crmURL p="civicrm/sepa/listgroup" q="group_id=$group_id"}" class="button_view" title="test" icon="fa-info"}{ts domain="org.project60.sepa"}Contributions{/ts}{/crmButton}
```
jensschuppe, line 111: `{crmButton}` doesn't need the `{crmURL}` tag, it has its own `p` attribute for a CiviCRM path.

jensschuppe, line 111: The `title` attribute seems off.

templates/CRM/Sepa/Page/DashBoard.tpl
```diff
@@ -105,27 +107,26 @@
...
       {if $group.status == 'open'}
         {if $can_batch}
           {if $group.submit == 'missed'}
-            <a href="{crmURL p="civicrm/sepa/closegroup" q="group_id=$group_id&status=missed"}" class="button button_close">
+            {crmButton href="{crmURL p="civicrm/sepa/closegroup" q="group_id=$group_id&status=missed"}" class="button_close" title="{ts domain="org.project60.sepa"}Close and Submit{/ts}" icon="fa-paper-plane"}{ts domain="org.project60.sepa"}Close and Submit{/ts}{/crmButton}
```
jensschuppe, line 115: Translations within HTML attributes need the `escape="htmlattributes"` parameter.

sepa.php
```diff
@@ -608,3 +599,10 @@ function sepa_civicrm_xmlMenu(array &$files): void {
     $files[] = $file;
   }
 }
+
+function sepa_civicrm_coreResourceList(&$list, $region) {
```
jensschuppe, line 603: Let's add type hints and the conventional docblock to make PHPStan happy.

jensschuppe (changes requested):
Please have a look at the PHPStan and PHP_CodeSniffer issues related to the files affected in this PR.

css/sepa.css
knottheory: This adresses the table cell containing the buttons. Setting the display property to flex aligns the buttons horizontally instead of vertical. The overall screen space the buttons take up is reduced.
Nonetheless the selector could be more descriptive.

templates/CRM/Sepa/Page/DashBoard.tpl
knottheory: Thanks for pointing it out I have updated the buttons accordingly.

css/sepa.css
knottheory: I have updated the class name following BEM conventions and renamed it to sepa_dashboard__button_group to make more descriptive.

templates/CRM/Sepa/Page/DashBoard.tpl
knottheory: I have added the crmScpoe tag and removed domain attributes from translation tags.

templates/CRM/Sepa/Page/DashBoard.tpl
knottheory: I fixed the title attribute.

sepa.php
knottheory: The type hints have been added.

css/sepa.css
jensschuppe: I'd suggest adding
```css
.sepa_dashboard .sepa_dashboard__button_group .button {
    height: fit-content;
}
```
for buttons not growing to the table row's height (which is the case when not using *RiverLea*-based themes such as the traditional *Greenwich*) and
```css
gap: var(--crm-flex-gap);
```
for the `.sepa_dashboard .sepa_dashboard__button_group` to retain a margin between buttons which is being set to `0` by *Riverlea* for some reason.

jensschuppe:
This needs a rebase onto current `master`.

knottheory:
@jensschuppe I rebased onto the master and it should be able to merge now.

PR #824, opened 2026-08-19
Batching: do not create a second FRST installment when an open group is overdue

### LINKED ISSUE #823: Batching creates a second FRST installment when an earlier FRST group was left open past its date
Follow-up to #634, which noted that "legitimate FRST contributions still will be collected duplicate in this case". We hit it in production; PR follows.

**Reproduce:** create an RCUR mandate, run the FRST batching (pending first installment in `TXG-…-FRST-<date1>`), leave that group open until `date1` has passed, run the FRST batching again.

**Result:** a second pending first installment in `TXG-…-FRST-<date2>`; the first one stays in the old open group. Closing both collects the mandate twice.

**Cause:** `Batching::updateRCUR` STEP 3 matches an existing installment only by `DATE(receive_date) = <new collection date>`, so STEP 4 creates another one and `syncGroups()` only moves the new one. RCUR mandates are anchored to the first collection; FRST mandates have no anchor.

jfilter:
Fixes #823.

- FRST-STEP 3b in `Batching::updateRCUR`: a pending first installment in an open FRST group of the same creditor is re-dated to the new collection date and reused, so `syncGroups()` moves it and the emptied group is cleaned up. Several candidates → warning, mandate skipped for this run. On-hold installments are not re-dated.
- `hook_civicrm_check` warning for open, non-empty groups past their submission deadline.
- Tests for the overdue-group case, the ambiguous case and the status check; one paragraph in `docs/usage.md`.

Re-dating is consistent with `Group::received()` and STEP 4 and keeps the date-based lookup idempotent. RCUR is unchanged: a missed RCUR rate may legitimately be collected later.

### DIFF
--- a/CRM/Sepa/Logic/Batching.php
+++ b/CRM/Sepa/Logic/Batching.php
@@ -231,6 +231,72 @@ class CRM_Sepa_Logic_Batching {
       /** @var array<int, array{id: int, contribution_recur_id: int, "sepa_contribution.is_on_hold": ?bool}> $existingContributionsByRecurId */
     }
 
+    // FRST-STEP 3b: reuse a pending first installment sitting in an open group with
+    // another date (e.g. an overdue group) instead of creating a second one.
+    if ('FRST' === $mode) {
+      foreach ($rcontribIdsByCollectionDate as $collectionDate => $rcontribIds) {
+        $unmatchedRecurIds = array_values(array_diff($rcontribIds, array_keys($existingContributionsByRecurId)));
+        if ([] === $unmatchedRecurIds) {
+          continue;
+        }
+        /** @var list<array{id: int, contribution_recur_id: int, "sepa_contribution.is_on_hold": ?bool, "txg.id": int}> $pendingInstallments */
+        $pendingInstallments = Contribution::get(FALSE)
+          ->addSelect('id', 'contribution_recur_id', 'sepa_contribution.is_on_hold', 'txg.id')
+          ->addJoin('SepaContributionGroup AS ctxg', 'INNER', NULL, ['ctxg.contribution_id', '=', 'id'])
+          ->addJoin('SepaTransactionGroup AS txg', 'INNER', NULL, ['txg.id', '=', 'ctxg.txgroup_id'])
+          ->addWhere('contribution_recur_id', 'IN', $unmatchedRecurIds)
+          ->addWhere('contribution_status_id:name', '=', 'Pending')
+          ->addWhere('payment_instrument_id', 'IN', $paymentInstrumentIds)
+          ->addWhere('txg.type', '=', 'FRST')
+          ->addWhere('txg.sdd_creditor_id', '=', $creditorId)
+          ->addWhere('txg.status_id', '=', $groupStatusIdOpen)
+          ->addWhere('is_test', 'IS NOT NULL')
+          ->addOrderBy('id')
+          ->execute()
+          ->getArrayCopy();
+        /** @var array<int, list<array{id: int, contribution_recur_id: int, "sepa_contribution.is_on_hold": ?bool, "txg.id": int}>> $pendingByRecurId */
+        $pendingByRecurId = [];
+        foreach ($pendingInstallments as $installment) {
+          $pendingByRecurId[$installment['contribution_recur_id']][] = $installment;
+        }
+        foreach ($pendingByRecurId as $recurId => $installments) {
+          if (count($installments) > 1) {
+            // ambiguous: neither guess nor add another one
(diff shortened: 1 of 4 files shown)


CRM/Sepa/Logic/Batching.php
```diff
@@ -231,6 +231,72 @@ public static function updateRCUR(
...
+              . "installments in open groups ($where), skipped."
+            );
+            foreach ($mandatesByCollectionDateAndFinancialTypeId[$collectionDate] as &$mandates) {
+              $mandates = array_values(array_filter(
+                $mandates,
+                /** @param array<string, mixed> $m */
+                fn(array $m) => $m['entity_id'] !== $recurId
+              ));
+            }
+            unset($mandates);
```
dontub, line 280: What's the aim of this `unset()`? Currently I'd think that this line doesn't affect the result.

CRM/Sepa/Logic/Batching.php
```diff
@@ -231,6 +231,72 @@ public static function updateRCUR(
...
+            continue;
```
dontub: This `continue` at the end of the loop body has no influence.

CRM/Sepa/Logic/Batching.php
```diff
@@ -231,6 +231,72 @@ public static function updateRCUR(
...
+    // FRST-STEP 3b: reuse a pending first installment sitting in an open group with
+    // another date (e.g. an overdue group) instead of creating a second one.
+    if ('FRST' === $mode) {
+      foreach ($rcontribIdsByCollectionDate as $collectionDate => $rcontribIds) {
+        $unmatchedRecurIds = array_values(array_diff($rcontribIds, array_keys($existingContributionsByRecurId)));
+        if ([] === $unmatchedRecurIds) {
+          continue;
+        }
+        /** @var list<array{id: int, contribution_recur_id: int, "sepa_contribution.is_on_hold": ?bool, "txg.id": int}> $pendingInstallments */
+        $pendingInstallments = Contribution::get(FALSE)
```
dontub, line 243: Would it be better to name it `$pendingFrstInstallments`?

CRM/Sepa/Logic/Batching.php
```diff
@@ -231,6 +231,72 @@ public static function updateRCUR(
...
+          ->addWhere('payment_instrument_id', 'IN', $paymentInstrumentIds)
+          ->addWhere('txg.type', '=', 'FRST')
+          ->addWhere('txg.sdd_creditor_id', '=', $creditorId)
+          ->addWhere('txg.status_id', '=', $groupStatusIdOpen)
+          ->addWhere('is_test', 'IS NOT NULL')
+          ->addOrderBy('id')
+          ->execute()
+          ->getArrayCopy();
+        /** @var array<int, list<array{id: int, contribution_recur_id: int, "sepa_contribution.is_on_hold": ?bool, "txg.id": int}>> $pendingByRecurId */
+        $pendingByRecurId = [];
```
dontub, line 258: Would it be better to add *frst* to the variable name here, too?

CRM/Sepa/Logic/Batching.php
```diff
@@ -231,6 +231,72 @@ public static function updateRCUR(
...
+        foreach ($pendingInstallments as $installment) {
+          $pendingByRecurId[$installment['contribution_recur_id']][] = $installment;
+        }
+        foreach ($pendingByRecurId as $recurId => $installments) {
+          if (count($installments) > 1) {
```
dontub, line 263: The line `$existingContributionsByRecurId[$recurId] = $installment;` below would prevent new FRST contributions being created, if there's already one pending FRST contribution, and the existing contribution would be moved to a new transaction group. So `count($installments) > 1` can actually only be true in cases before the code was added. In such a case the filtering of `$mandates` below would result in no new FRST contribution being created and the existing FRST contributions might be removed from there current groups, depending on whether a mandate with the same collection date remains in `$mandatesByCollectionDateAndFinancialTypeId`. Is my understanding correct or have I missed something?

CRM/Sepa/Logic/Batching.php
```diff
@@ -231,6 +231,72 @@ public static function updateRCUR(
...
+            // ambiguous: neither guess nor add another one
+            $where = implode(', ', array_map(
+              fn(array $i) => "contribution {$i['id']} in group {$i['txg.id']}",
```
dontub, line 266: I'm not sure if it's a good idea to shorten `$installment` to `$i`. I think `$installment` would result in better readability.

dontub (changes requested):
Thanks for the PR @jfilter! Can you have a look at my comments and questions?

PR #827, opened 2026-09-08
CRM_Sepa_Logic_Queue_Update: Fix argument type

dontub:
`CRM_Sepa_Logic_Queue_Update` expects the creditor ID as int, but got it as string. The APIv3 call to get the creditor IDs is replaced by an APIv4 call.


jensschuppe (approved):

jensschuppe:
PHPStan and PHPUnit errors don't seem related.

PR #830, opened 2026-09-14
Fix BugReproductionTest

dontub:


jensschuppe (approved):
Adding context: the test failed because of incomplete assumptions on dates for batching groups, i.e. July and August both having 31 days caused the chosen intervals of 30 days to not match the installment cycle. By selecting `month` as the unit instead, the test should pass independent from actual dates. The test itself was not about dates in the first place, though.

